PR #15, opened 2016-10-12
Install note for Windows devs

capajon:
I'm not 100% sure of everything I did to get it working on Windows but I know it at least included the items I listed in the note, which might help someone in the future.


dlockhart:
Actually don't merge this, I noticed on Jeff's Mac that I needed to do the same thing -- I'll fix this a different way.

dlockhart:
I created #16 for me to do this

PR #17, opened 2016-10-12
fixes #16: installing ruby deps as part of NPM install

### LINKED ISSUE #16: ruby dependencies need to get installed
The `pygments` and `jekyll-minibundle` don't get installed automatically the first time you set this up. Should either find a way to do that automatically, or include it in the install instructions.

dlockhart:
@capajon seem OK?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ rvm:
 before_install:
   - nvm install stable
 install:
-  - gem install jekyll pygments.rb jekyll-minibundle
   - nvm use stable
   - npm install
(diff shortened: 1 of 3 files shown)


capajon:
Looks good.  That will help.

 We should probably include a note though that python and pygments and ruby devkit may be required on Windows (I'm not sure if they are, don't really want to start it from scratch to prove it either way though lol).

dlockhart:
Yeah agreed, feel free to re-open your PR -- sorry about that.

PR #18, opened 2016-10-12
d2l-button JSFiddle example

capajon:
Got our components going in JSFiddle.  What do you think?  I like the idea of someone being able to play with it live inline.  These pages could even form the basis of the demo for next Fusion actually!

Anyway, certainly can be changed - I didn't put a lot of thought into any supporting text and what not.

Take a look.  I can easily add this for the other components if this one looks good.

### DIFF
--- a/_components/buttons.md
+++ b/_components/buttons.md
@@ -72,6 +72,6 @@ Then import the components you need:
 
 ### Fiddle with it!
 
-<script async src="//jsfiddle.net/capa/z34s1zdq/embed/html,result/"></script>
+<script async src="//jsfiddle.net/capa/z34s1zdq/embed/html,result/?bodyColor=f9fafb"></script>
 
 {% include github.html name=page.github %}
--- a/_components/buttons.md
+++ b/_components/buttons.md
@@ -70,4 +70,8 @@ Then import the components you need:
   href="bower_components/d2l-button/d2l-floating-buttons.html">
 {% endexample %}
 
+### Fiddle with it!
+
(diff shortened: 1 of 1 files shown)


capajon:
Oh, not sure if we want a generic jsfiddle account so it isn't tied to anyone personally.

capajon:
@dlockhart Any change to review?  Worth adding fiddles for the other components?

dlockhart:
@capajon hey, sorry. This is super neat! I'm still on the fence about the value it adds for an individual component compared to the code samples. What might be cool is to use it as a generic playground (like a separate section in the docs) to experiment with ALL the components in one place. Have you tried any other sandboxes like jsbin? It's a little more popular these days...

capajon:
@dlockhart - I also wondered if it is a bit redundant but thought it was still valuable as a way to sort of lower the barrier to trying out our components even further.  Really the examples could almost use this approach and kill two birds with one stone, but there are some downsides (like relying on polygit.org to not die).

Pulling it into one area could be cool though - and would work as a demo alternative for things like what we do at the extensibility lab.  

Re. JSBin - they are all so similar these days that I didn't put a lot of thought into that.  But it looks like JSBin's embedded examples can be edited inline, which sounds a lot better for this, so I'll give it a try. 

So I'll go ahead when I have time in between stories and create a new section and see what it looks like.

dlockhart:
Sounds good, thanks for trying this out!

dlockhart:
Closing due to old age.

PR #25, opened 2017-01-11
Add some table documentation

awikkerink:


_components/tables.md
```diff
@@ -0,0 +1,237 @@
+---
+title: Tables
+permalink: /components/tables/
+layout: component
+github: BrightspaceUI/table
+bower: d2l-table
+---
+A responsive styled table.
+
+To create a table, extend the native HTML `<table>` element with with the `is="d2l-table"` custom element and optionally surround it with a `<d2l-table-wrapper>` element.
```
dlockhart: Under what conditions should I or should I not optionally surround it in `<d2l-table-wrapper>`?

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+
+{% example %}
+<d2l-table-wrapper>
+	<table is="d2l-table">
+		<thead>
+			<th>
+				<button is="d2l-table-col-sort-button">Ascending</button>
```
dlockhart: Maybe a small blurb about the three sorting attributes which can be applied?

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+			<td>321</td>
+		</tr>
+		</tbody>
+	</table>
+</d2l-table-wrapper>
+{% endexample %}
+
+## Responsive
+
+The d2l-table is automatically responsive
```
dlockhart: Missing period at the end of the sentence. Suggestion:

"The table is responsive. When a table is wider than a user's screen, arrows will appear which allow the user to scroll the table back and forth."

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+
+{% example %}
+<d2l-table-wrapper style="max-width: 200px;">
```
dlockhart: For the responsive demo, since there isn't anything code-related the developer needs to do, I'd suggest not wrapping it in `{%example %}` but just putting it inline. That way it'll just be an interactive demo instead of a code sample + demo.

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+</d2l-table-wrapper>
+{% endexample %}
+
+## No column border
+
+To disable column borders, add the `no-column-border` attribute to the `table` element
+
+{% example %}
+<d2l-table-wrapper>
+	<table is="d2l-table" no-column-border>
```
dlockhart: The full table example is a lot of HTML to keep showing -- maybe for this case just have enough table markup to demonstrate what this section is introducing?

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+			<td>321</td>
+		</tr>
+		</tbody>
+	</table>
+</d2l-table-wrapper>
+{% endexample %}
+
+## Selectable
+
+To select a row, add the `selected` attribute to the `tr` element. To make a table rows respond to hovering, add the `selectable` attribute to the `table` element.
```
dlockhart: Seems like you'd always want to use these 2 attributes in conjunction? Suggestion:

"To hint that rows are selectable with a hover effect, add the `selectable` attribute to the `table` element. To set a row as being selected, add the `selected` attribute to its `tr` element."

dlockhart: Not sure if it's worth giving an example of how they'd add checkboxes and wire them all up to set the right attributes... could get complicated though!

_components/tables.md
```diff
@@ -0,0 +1,237 @@
...
+			<td>456</td>
+			<td>321</td>
+			<td>321</td>
+		</tr>
+		</tbody>
+	</table>
+</d2l-table-wrapper>
+{% endexample %}
+
+## Styling
```
dlockhart: Is custom styling something we expect most people to need to do often? If not, this type of documentation might belong more in the GitHub README.

PR #32, opened 2017-03-01
Dbatiste/dropdown doc

dbatiste:
@dlockhart : I was chatting with Avis (re their plagiarism app) yesterday and noticed that this common component with missing from our docs, so I took a stab at adding this.  I didn't document absolutely everything - I tried to stick with what we expect consumers to do.  Any feedback is much appreciated.  :)

### DIFF
--- a/_components/dropdowns.md
+++ b/_components/dropdowns.md
@@ -56,7 +56,7 @@ The `d2l-dropdown-context-menu` simplifies your mark-up and provides consistency
 
 ## Dropdown Menu Content
 
-The `d2l-dropdown-menu` is a special content component intended the for `d2l-menu` component.
+The `d2l-dropdown-menu` is a special content component intended the for [d2l-menu](/components/menus/) component.
 
 {% example %}
 <d2l-dropdown-button text="Open!" primary>
--- a/_components/dropdowns.md
+++ b/_components/dropdowns.md
@@ -11,7 +11,7 @@ To create a dropdown, use the `d2l-dropdown` element. Provide an opener element 
 
 {% example %}
 <d2l-dropdown>
-  <button class="d2l-dropdown-opener">Open!</button>
+  <button is="d2l-button" class="d2l-dropdown-opener">Open!</button>
   <d2l-dropdown-content>
       Some content...
   </d2l-dropdown-content>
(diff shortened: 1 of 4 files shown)


dlockhart:
Working version of the PR here:
http://ui-dev.developers.brightspace.com.s3-website-us-east-1.amazonaws.com/pr/32/

_components/dropdowns.md
```diff
@@ -0,0 +1,108 @@
...
+  </d2l-dropdown-content>
+</d2l-dropdown>
+{% endexample %}
+
+### Dropdown Content Options
+
+The `d2l-dropdown-content` component supports a number of optional attributes to configure its appearance and behavior.
+
+{% example demo: false %}
+<d2l-dropdown-content min-width="100" max-width="500" vertical-offset="0" no-padding render-content>
```
dlockhart: I worry a bit that people will just copy-paste this since it's the first thing they see without understanding what all the optional attributes do and whether they actually need them. Suggestion: don't include any optional attributes there in this first example, but just list them out after with descriptions of what they do and when to use them.

dlockhart: Hmm, I guess this is the second time they've seen it... maybe nevermind :)

dlockhart:
Oh one suggestion: since they're used so often together, it might be nice to create some links between the new dropdown documentation and the existing menu stuff.

_components/dropdowns.md
dbatiste: Cool - will do.

dbatiste: Ah k - yeah, I guess I will leave this as-is, since it shows how it would be formatted and without units.

_components/dropdowns.md
```diff
@@ -0,0 +1,108 @@
...
+<d2l-dropdown-context-menu text="Open!">
+    <d2l-dropdown-content>
+        Some content...
+    </d2l-dropdown-content>
+</d2l-dropdown-context-menu>
+{% endexample %}
+
+## Dropdown Menu Content
+
+The `d2l-dropdown-menu` is a special content component intended the for [d2l-menu](/components/menus/) component.
```
dbatiste: Updated to provide a link to `d2l-menu`.  @dlockhart : were you thinking of something more to link the two?

PR #35, opened 2017-06-05
Add example for align and boundary properties

awikkerink:


dlockhart:
Looks good, although you may want to wait until the new functionality is in a dropdown version and upgrade to that version at the same time as this PR.

PR #58, opened 2020-06-19
Component status table: tweak style & messages

margaree:
Switched from status indicator to icon + link (if readme link) or icon + text (if no link) in order to make it more clear that there is a link present

### DIFF
--- a/src/component-list.js
+++ b/src/component-list.js
@@ -2,7 +2,6 @@ import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import '@brightspace-ui/core/components/inputs/input-search.js';
 import '@brightspace-ui/core/components/link/link.js';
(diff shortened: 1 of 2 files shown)


src/component-list.js
```diff
@@ -94,5 +115,24 @@ export class DesignSystemComponentList extends LitElement {
...
+				break;
+			case 'success':
+				icon = 'tier1:check';
+				break;
+		}
+		const stateText = readme
+			? html`<d2l-link small href="${readme}">${componentStateInfo.text}</d2l-link>`
+			: html`<div class="d2l-body-small d2l-design-system-component-list-info-text">${componentStateInfo.text}</div>`;
+		return html`<d2l-icon data-state="${componentStateInfo.state}" icon="${icon}"></d2l-icon>${stateText}`;
+	}
```
dlockhart, line 136: Nit: could move outside the component class.

PR #61, opened 2020-06-23
switching to use site structure data to lay out the site

dlockhart:
Alright, finally got something not-crazy-big that didn't break everything you had working. :)

Basically this introduces a `structure.js` that outlines the structure of the site's section. We could add another level of navigation at the top using this later if we want. Each entry in the structure is the same -- it has a name, a type, some custom data and possibly children.

The name of each entry plus all its descendants is how the URL gets crafted.

The type can currently be:
- `custom`: which we can use for custom Lit views like the Component Status page
- `container`: basically just a hint to make the URL the URL of the first child
- `component`: a hint so we load the component view
- `??`: future types like Markdown maybe

I think this is enough so that I can try to implement Markdown page types that can be generated from component READMEs (or a combination of multiple READMEs) or just pointed at local Markdown files in the docs repo (which we could use for the Welcome page for example).

### DIFF
--- a/data/structure.js
+++ b/data/structure.js
@@ -1,0 +1,131 @@
+import { default as components } from '../data/components.js';
+
+const siteStructure = [
+	{
+		name: 'Welcome',
+		type: 'custom'
+	},
+	{
+		name: 'Component Status',
+		type: 'custom'
+	},
+	{
+		name: 'Actions',
+		type: 'container'
+	},
+	{
+		name: 'Feedback',
+		type: 'container'
+	},
+	{
+		name: 'Overlay',
+		type: 'container'
+	},
+	{
+		name: 'Navigation',
+		type: 'container'
+	},
+	{
+		name: 'Structure',
+		type: 'container'
+	},
+	{
+		name: 'Forms',
+		type: 'container'
+	},
+	{
+		name: 'Other',
+		type: 'container'
+	}
+];
+components.forEach((component) => {
+	let cat = siteStructure.find((c) => c.name === component.type);
+	if (!cat) {
+		cat = { name: component.type };
+		siteStructure.push(cat);
+	}
+	cat.children = cat.children || [];
+	cat.children.push(convertComponentToItem(component));
+});
+
+function convertComponentToItem(component) {
+	const item = { name: component.name, type: 'component', children: [], data: { tagName: component.tag } };
+	if (component.childComponents) {
+		item.type = 'container';
+		item.children = component.childComponents.map((c) => convertComponentToItem(c));
+	}
+	return item;
+}
+
+function makeUrlFriendly(name) {
+	return name.replace(/\s/g, '-').toLowerCase();
+}
+
+export function findItemFromPath(path) {
+	function search(items, parents) {
+		for (let i = 0; i < items.length; i++) {
+			const item = items[i];
+			parents.push(item);
+			const itemPath = `/components/${parents.map((p) => makeUrlFriendly(p.name)).join('/')}`;
+			if (itemPath === path) {
+				return item;
+			}
+			if (item.children) {
+				const result = search(item.children, parents);
+				if (result !== null) {
+					return result;
+				}
+			}
+			parents.pop();
+		}
+		return null;
+	}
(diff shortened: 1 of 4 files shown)


data/structure.js
```diff
@@ -0,0 +1,131 @@
...
+	{
+		name: 'Forms',
+		type: 'container'
+	},
+	{
+		name: 'Other',
+		type: 'container'
+	}
+];
+components.forEach((component) => {
```
dlockhart, line 41: Currently this just smushes your existing `components.js` file into here, transforming the component entries into site structure entries. We may choose to just embed these right in the structure itself later.

src/app.js
```diff
@@ -4,20 +4,64 @@ import './welcome.js';
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/icons/icon.js';
 import { css, html, LitElement } from 'lit-element';
+import { findItemFromPath, getItemPath, siteStructure } from '../data/structure.js';
 import { classMap } from 'lit-html/directives/class-map.js';
-import { default as components } from '../data/components.js';
 import { linkStyles } from '@brightspace-ui/core/components/link/link.js';
 import page from 'page';
 
+function buildSidebarItem(item, parents, currentPath) {
```
dlockhart, line 12: I refactored this out to try and reuse some rendering, but I think we should just pull it right out into a sidebar component. Didn't want to spend too much time on this as I'm sure once design has a go at this they'll switch things up a lot!

src/app.js
```diff
@@ -4,20 +4,64 @@ import './welcome.js';
...
+
 export class DesignSystem extends LitElement {
 	static get properties() {
 		return {
-			_categories: { type: Object },
-			_component: { type: String },
-			_currentView: { type: String },
-			_shownCategory: { type: String },
-			_shownComponent: { type: String },
-			_shownNested: { type: String }
```
dlockhart, line 20: I was able to replace all this tracking with just the current path.

src/app.js
```diff
@@ -178,68 +167,23 @@ export class DesignSystem extends LitElement {
...
-		const componentName = context.params['component'];
-		const filtered = components.filter((component) =>  component.tag === componentName);
-		this._component = JSON.stringify(filtered[0]);
-		this._shownCategory = filtered[0].type;
-		this._shownComponent = filtered[0].name;
-		this._shownNested = '';
-	}
-
 	_installRoutes() {
+		page.base('/components');
```
dlockhart, line 171: This command essentially nests the entire site under `/components`. If we introduce other top-level navigation items, we can remove this.

src/component.js
```diff
@@ -42,9 +42,7 @@ function _getTable(componentInformation, title, hasTypeAndDefault) {
 export class DesignSystemComponent extends LitElement {
 	static get properties() {
 		return {
-			component: { type: String },
-			_componentInfo: { type: Object },
-			_componentName: { type: String }
+			tagName: { type: String, attribute: 'tag-name', reflect: true }
```
dlockhart, line 45: Seemed easier to just pass the tagName down to the component view and then have it fetch its data.

src/app.js
```diff
@@ -98,59 +142,10 @@ export class DesignSystem extends LitElement {
...
-				if (component.childComponents) {
-					const nestedChildren = component.childComponents.map((childComponent) => {
-						const classes = {
-							'd2l-link': true,
-							'd2l-link-small': true,
-							'd2l-design-system-link-selected': this._shownComponent === childComponent.name
-						};
-						return html`<li><a class="${classMap(classes)}" href="/components/${component.name}/${childComponent.tag}">${childComponent.name}</a></li>`;
-					});
-					return html`<li class="d2l-design-system-nested">
```
margaree, line 127: Nit: this css https://github.com/BrightspaceUI/documentation/blob/32a12a9d79bf1ee0c87b03a83f157d1e81c23388/src/app.js#L128 could be cleaned up because of this class no longer being used

dlockhart: Oops I just missed what that class was doing -- added it back in so those items are lined up like you had them.

PR #62, opened 2020-06-24
Component data updates

margaree:
- added the custom-components.json file generated from facet-filter-sort (only did the one for now to see if this seems like a good path)
- added missing components from core to components.js
- updated statuses in components.js
- did some parsing of type for the attributes table


data/custom-elements-facet-filter-sort.json
```diff
@@ -0,0 +1,634 @@
+{
```
dlockhart, line 1: Probably didn't intend to check this in?

margaree: So I did because we don't have facet-filter-sort as a dependency so we can't necessarily get at its custom-elements.json file, and I wasn't sure if it was better to have the file here (where it could go out of date, but less dependencies), or add facet-filter-sort as a dependency. Thoughts?

dlockhart: I'm OK adding it as a devDependency, since it'll need to exist anyway if we're going to pull in its README and have a live demo.

PR #63, opened 2020-06-25
Component: Add column to attributes table for setting the attribute in future demo

margaree:
Attributes with type `array` and `object` are not handled as I wasn't totally sure what to do there. Could also do a input-text which contains empty `{}` or `[]` for the user to fill in.

Example (from d2l-tooltip):
Screen Shot 2020-06-25 at 10 35 47 AM

### DIFF
--- a/src/component-attribute-table.js
+++ b/src/component-attribute-table.js
@@ -18,7 +18,7 @@ const validTypes = [
 export class DesignSystemComponentAttributeTable extends LitElement {
 	static get properties() {
 		return {
-			attributes: { type: Array, reflect: true }
+			attributes: { type: String, reflect: true }
 		};
 	}
 
@@ -45,7 +45,8 @@ export class DesignSystemComponentAttributeTable extends LitElement {
 
 	render() {
 		if (!this.attributes) return;
-		const rows = this.attributes.map((info) => {
+		const attributes = JSON.parse(this.attributes);
+		const rows = attributes.map((info) => {
 			const infoDefault = info.default ? info.default.replace(/\\"/g, '') : null;
 			let demoType = info.type;
 			let demoValue = null;
@@ -136,8 +137,7 @@ export class DesignSystemComponentAttributeTable extends LitElement {
 						aria-label="Valid values"
 						@change="${this._onSelectChange}"
 						class="d2l-input-select"
-						data-name="${attributeName}"
-						type="select">
+						data-name="${attributeName}">
 						${optsHtml}
 					</select>
 				`;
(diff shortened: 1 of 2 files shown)


src/component.js
```diff
@@ -1,35 +1,125 @@
...
+					@change="${changeFunc}"
+					data-name="${attributeName}"
+					label="${attributeName}"
+					label-hidden
+					value="${ifDefined(value)}">
+				</d2l-input-text>
+			`;
+		default: {
+			// the case of an array of strings
+			const optsHtml = type.replace(/'/g, '').split('|').map(opt => html`<option>${opt}</option>`);
```
dlockhart, line 90: Looks like you're not selecting the option that matches `value`?

src/component.js
```diff
@@ -1,35 +1,125 @@
...
+			return html`
+				<label>
+					<span class="d2l-input-label">Valid string values</span>
```
dlockhart, line 93: Could probably make this an `aria-label` on the `select`? Like technically it's a string and these are the only valid values... but really it's an enum.

src/component.js
```diff
@@ -89,5 +186,11 @@ export class DesignSystemComponent extends LitElement {
 		`;
 	}
 
+	_onChange(e) {
+		const attributeName = e.target.getAttribute('data-name');
+		const attributeValue = e.target.value || e.target.hasAttribute('on');
```
dlockhart, line 191: If we end up firing an event that contains all the details, we'll likely want to include the attribute name (which you have), the type and then depending on the type have the value either be a boolean, string, number, etc. So I'm wondering if instead of having a generic `_onChange` if it'd be easier to wire up a different handlers depending on the component used and then have them fire the event. 🤷

src/component.js
margaree: Yeah - I'd wanted to leave that as text in order to prevent confusion as to why it's a string type but then you can only choose one of these values.

src/component-attribute-table.js
```diff
@@ -0,0 +1,185 @@
...
+	'boolean',
+	'number',
+	'object',
+	'string'
+];
+
+export class DesignSystemComponentAttributeTable extends LitElement {
+	static get properties() {
+		return {
+			attributes: { type: String, reflect: true }
```
dlockhart, line 21: Other option here is to give it the `tag-name="d2l-button"` and then this can include `component-doc-details.js` and pull the right info out. Would definitely make it easier for me to consume this component later from a README when I don't necessarily have all the attributes.

PR #69, opened 2020-06-26
Add interactive demo

margaree:


src/component-attribute-table.js
```diff
@@ -19,12 +21,15 @@ const validTypes = [
 export class DesignSystemComponentAttributeTable extends LitElement {
 	static get properties() {
 		return {
-			tagName: { type: String, attribute: 'tag-name', reflect: true }
+			code: { type: String, reflect: true },
+			hideDemo: { type: Boolean, attribute: 'hide-demo', reflect: true },
```
margaree, line 25: `hideDemo` is temporary while we still have the non-markdown pages since the demo does not work but we still want to include the attributes table

src/component-attribute-table.js
```diff
@@ -69,7 +84,14 @@ export class DesignSystemComponentAttributeTable extends LitElement {
 				</d2l-tr>
 			`;
 		});
+		const demo = !this.hideDemo ? html`
+			<d2l-design-system-interactive-demo
```
margaree, line 88: I nested the demo in the table component because that seemed like the easiest way to pass the changed attributes and that seemed a bit complicated if both were in `generate.js`

src/component-attribute-table.js
```diff
@@ -88,11 +110,13 @@ export class DesignSystemComponentAttributeTable extends LitElement {
...
-			{ bubbles: true, composed: false, detail: details }
-		));
-		console.warn(`ATTRIBUTE CHANGE: ${JSON.stringify(details)}`);
+		const tempAttributes = this._changedAttributes;
+		this._changedAttributes = {};
+		tempAttributes[details.name] = {
+			value: details.value,
+			type: details.type
+		};
+		this._changedAttributes = tempAttributes;
```
margaree, line 119: I did the modification of attributes through reassignment so that it would trigger a render

src/interactive-demo.js
```diff
@@ -0,0 +1,86 @@
...
+					} else {
+						component.setAttribute(key, attributes[key].value);
+					}
+				});
+				this.shadowRoot.querySelector('d2l-demo-snippet').forceCodeUpdate();
+			}
+		});
+	}
+
+	async _getCode() {
```
margaree, line 63: This part feels a bit fragile - basically it removes the imports and creates a new `script` and then adds those imports so that they actually get imported

dbatiste:
I'm sorta interested in where this is heading.  One of the things I was considering for this inspiration spring was - updating the demo-snippet to enable live-editing, with syntax highlighting.  Maybe useful here...?

Some components require more than one import - for example, menus require importing the menu and also menu item components used in the menu.  Dropdowns are similar, and further require importing menus and even tabs for those use-cases.  Does this PR set up for those scenarios?

src/component-attribute-table.js
dlockhart: I think we'll likely want to reverse this around at some point... like have `<d2l-design-system-interactive-demo code="whatever">` and then it renders the demo snippet and the attribute table? But anything is good for now to get the demo working. :)

margaree:
Awesome - I think live editing would be useful here.
I did test with multiple imports and multiple components and that does work

src/component-attribute-table.js
margaree: Yeah that would make the most sense in the long run.

dlockhart:
Where I think this is heading:
- Using an IFRAME to host the demo snippet -- that way we can add a slider to resize it and have the actual responsive rules kick in. It also sandboxes the snippet nicer from the host page.
- Better automatic detection of what element is actually being demo'd here, and then we can automatically display the interactive attribute table and populate it with the correct values based on what's set. If we use a `<template>` we can literally just walk the DOM to find the first element and then based on changes to the attributes we can keep that in all in sync easily.
- Having a dropdown (or other UI mechanism) so that you could set up many different example scenarios and be able to switch between them. Example: simple menu, nested menu, menu with checkboxes. Or: dropdown with different openers. That way the page can be succinct and not drag on with examples each with their own giant attribute table.

I'm not sure where live editing fits in here. If the user can type whatever they want, then the binding with the interactive attribute table becomes complicated. What happens if they use something that isn't imported? Maybe we have a separate "sandbox" area for playing around like this?

dbatiste:
Sounds great.

Re: finding first element and syncing attributes...  for something like below, the `d2l-dropdown-content` doesn't really make a lot of sense on it's own, only when it's used in combination with a dropdown opener, yes it's not the first custom element, and it has lots of properties that can be toggled.  Food for thought.
```
<d2l-dropdown>
    <d2l-button class="d2l-dropdown-opener">Open it!</d2l-button>
    <d2l-dropdown-content max-width="400">
        stuff
    </d2l-dropdown-content>
</d2l-dropdown>
```

Re: What happens if they use something that isn't imported?

Yeah that's an interesting scenario.  I was mostly thinking of the user changing attributes.  But if they decided to change a `d2l-menu-item` to a `d2l-menu-item-link` and it hadn't been imported then they'd end up having to write their own script block to import it - not amazing.  I think the dropdown with preconfigured options is good.  The sandbox might be interesting - but there is always StackBlitz too (I actually have a template/starter for `@brightspace-ui\core` that can be forked if people really want to mess around.

I've been thinking about the rich source code editing in the context of the HTML Editor since it's something that is desired, and potentially overlaps with demo snippets.  So... just wanted to raise awareness to that and any potential overlap.  :)

PR #70, opened 2020-06-26
extracting tag from code block

dlockhart:
Feel free to ignore this for now if it complicates your demo.

This change extracts the tag name to use from the code block itself, which lets us have multiple interactive components in a single Markdown file -- which is how Button's README is actually laid out.

I think the next logical step from here would be to extract any attributes that are set in the code sample and pass those somehow to your attribute table -- that way it's populated not just with the default values but with the actual values used in the demo.


dlockhart:
Just saw your PR -- I'll rebase mine on top, so closing for now.

PR #72, opened 2020-06-29
markdown page: use d2l-table for tables

margaree:


cli/generate.js
```diff
@@ -24,6 +24,12 @@ function createComponentRenderer(tagName) {
 				isExampleBlock = true;
 			}
 			return defaultRenderer.heading(text, level, raw, slugger);
+		},
+		table(header, body) {
+			const d2lBody = body.replace(/tr>/g, 'd2l-tr>').replace(/td>/g, 'd2l-td>');
+			const d2lHeader = header.replace(/tr>/g, 'd2l-tr>').replace(/th>/g, 'd2l-th>');
+
+			return `<d2l-table><d2l-thead>${d2lHeader}</d2l-thead><d2l-tbody>${d2lBody}</d2l-tbody></d2l-table>`;
```
dlockhart, line 32: Haha wow do we ever need to convert tables to Lit so we can just apply styles to native elements...

PR #93, opened 2021-04-08
[Component Catalog] Initial work to set up 11ty site

margaree:
This is based off of the work in the CC spikes, but pared-down to only include that parts necessary to get a basic working site with side navigation (i.e., nothing related to pulling in components or complex markdown parsing).

The main piece that I'm wondering if I did correctly is the rollup since I haven't set that up before. I modelled it off of BSI but don't know if there are pieces of that that I can simplify (it does work as is).


.eleventy.js
```diff
@@ -0,0 +1,24 @@
...
+	eleventyConfig.addFilter("cssmin", function(code) {
+		return new cleanCSS({}).minify(code).styles;
+	});
+
+	eleventyConfig.addPlugin(eleventyNavigationPlugin);
+	eleventyConfig.setUseGitIgnore(false);
+	eleventyConfig.setLibrary("md", markdownIt);
+
+	return {
+	  dir: { input: 'pages', output: '_site' }
```
dlockhart, line 22: Surprised the mixed spaces/tabs isn't failing the linter?

.eleventy.js
```diff
@@ -0,0 +1,24 @@
...
+module.exports = function(eleventyConfig) {
+	eleventyConfig.addPassthroughCopy("assets/js");
+
+	let options = {
+		html: false,
+		breaks: false,
+		linkify: true
+	};
+	const markdownIt = require('markdown-it')(options);
+	eleventyConfig.addFilter("cssmin", function(code) {
```
dlockhart, line 13: Should probably be single quotes? (again curious why the linter isn't catching this)

package.json
```diff
@@ -1,43 +1,37 @@
...
-    "start:build": "npm run generate && npm run build && es-dev-server --root-dir dist --app-index index.html --open --compatibility none",
-    "start": "npm run generate && es-dev-server --app-index index.html --node-resolve --open --watch",
-    "test": "npm run lint",
-    "test:watch": "karma start --auto-watch=true --single-run=false"
+    "build": "run-s clean build:*",
+    "build:eleventy": "cross-env ELEVENTY_ENV=production eleventy",
+    "build:rollup": "rimraf assets/js && rollup -c rollup.config.js",
+    "clean": "rimraf _site",
+    "lint": "npm run lint:eslint",
+    "lint:eslint": "eslint --ext .js . --ignore-path .gitignore",
```
dlockhart, line 10: Does the default `eslint` (no parameters) not work? We used to have to specify this kind of stuff but didn't think we needed to anymore.

rollup.config.js
```diff
@@ -1,12 +1,18 @@
...
+const config = createBasicConfig({
+	developmentMode: false,
+	legacyBuild: true,
+	outputDir: 'assets/js'
 });
+config.input = componentFiles;
+config.output[0].entryFileNames = '[name].js';
+config.output[0].chunkFileNames = '[name].js';
+config.output[1].entryFileNames = 'nomodule-[name].js';
+config.output[1].chunkFileNames = 'nomodule-[name].js';
```
dlockhart, line 16: We only needed to override all this stuff for BSI because we needed the output filenames to be predictable, since we manually import them in the monolith pages. In your case, you're running Rollup on your entire project, so it'll use whatever filesnames it wants and then swap those into the HTML files.

.eleventy.js
margaree: 🤦‍♀️ yeah that mixing of quotes is gross. I'll investigate the linter.

package.json
margaree: So it seems like it does need the ` . --ext .js` part still. Looks like the .eleventy.js file was not getting picked up because of starting with a `.` so I need to add a .eslintignore file to specifically call out including it.

dlockhart (approved):

PR #95, opened 2021-04-14
Improve markdown parsing

margaree:
Implements the following:
- Replaces image src with correct path for screenshots
- Replaces link with `d2l-link`
- Parses demos and wraps them in corresponding component wrapper (`d2l-component-catalog-interactive-demo` does not yet exist)
- Handles hidden content and other types of sections (wraps them in a div with corresponding class)
- Uses Daylight colours for syntax highlighting that is similar to GitHub
- Adds an example component to demonstrate some of these parsing abilities

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,9 +1,8 @@
 /* global require, module */
 const cleanCSS = require('clean-css');
 const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
-const syntaxHighlight = require('@11ty/eleventy-plugin-syntaxhighlight');
+const { escapeHtml } = require('markdown-it/lib/common/utils');
 
-// TODO: imports
 module.exports = function(eleventyConfig) {
 	eleventyConfig.addPassthroughCopy('assets/js');
 	eleventyConfig.addPassthroughCopy('pages/components/imported/screenshots');
@@ -49,11 +48,23 @@ module.exports = function(eleventyConfig) {
 	const defaultFenceRule = markdownIt.renderer.rules.fence;
 	markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
-		if (content.includes('<!-- docs: live demo -->'))
-			return `<d2l-design-system-interactive-demo>${content}</d2l-design-system-interactive-demo>`;
-		else if (content.includes('<!-- docs: demo -->'))
-			return `<d2l-demo-snippet>${content}</d2l-demo-snippet>`;
-		else
+		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->')) {
+			let script = '';
+			if (content.includes('script')) {
+				const copy = content;
+				const scriptBeforeAfter = copy.split('</script>\n');
+				let imports = '';
+				const importsArray = scriptBeforeAfter[0].split('<script type="module">\n')[1].split('\n');
+				importsArray.forEach((importUrl) => {
+					const importStatement = importUrl.replace(/^(.*[\\/])/g, 'import \'/assets/js/'); // replace everything before final backslash
+					imports += `${importStatement}\n`;
+				});
+				script = `<script type="module">${imports}</script>`;
+			}
+
+			if (content.includes('<!-- docs: live demo -->')) return `${script}<d2l-component-catalog-interactive-demo>${escapeHtml(content)}</d2l-component-catalog-interactive-demo>`;
+			else return `${script}<d2l-component-catalog-demo-snippet-wrapper>${escapeHtml(content)}</d2l-component-catalog-demo-snippet-wrapper>`;
+		} else
 			return defaultFenceRule(tokens, idx, options, env, slf);
 	};
 
@@ -74,7 +85,6 @@ module.exports = function(eleventyConfig) {
 	};
 
 	eleventyConfig.addPlugin(eleventyNavigationPlugin);
-	eleventyConfig.addPlugin(syntaxHighlight);
 	eleventyConfig.setUseGitIgnore(false);
 	eleventyConfig.setLibrary('md', markdownIt);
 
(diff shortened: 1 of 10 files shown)


.eleventy.js
```diff
@@ -1,20 +1,87 @@
...
+	markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
+		const content = tokens[idx].content;
+		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->')) {
+			let script = '';
+			if (content.includes('script')) {
+				const copy = content;
+				const scriptBeforeAfter = copy.split('</script>\n');
+				let imports = '';
+				const importsArray = scriptBeforeAfter[0].split('<script type="module">\n')[1].split('\n');
+				importsArray.forEach((importUrl) => {
```
margaree, line 58: For each import (since there could be a few for a demo like menu) split on a new line, replace the import page with `/assets/js` (where the file should be post-rollup)

.eleventy.js
```diff
@@ -1,20 +1,87 @@
...
+					const importStatement = importUrl.replace(/^(.*[\\/])/g, 'import \'/assets/js/'); // replace everything before final backslash
+					imports += `${importStatement}\n`;
+				});
+				script = `<script type="module">${imports}</script>`;
+			}
+
+			if (content.includes('<!-- docs: live demo -->')) return `${script}<d2l-component-catalog-interactive-demo>${escapeHtml(content)}</d2l-component-catalog-interactive-demo>`;
+			else return `${script}<d2l-component-catalog-demo-snippet-wrapper>${escapeHtml(content)}</d2l-component-catalog-demo-snippet-wrapper>`;
```
margaree, line 66: So I used this wrapper strategy because just sending the code to `d2l-demo-snippet` it seems like there is an attempt to execute the original import, which doesn't work, and results in a console error.

.eleventy.js
```diff
@@ -1,20 +1,87 @@
...
+	};
+
+	const defaultTextRule = markdownIt.renderer.rules.text;
+	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
+		const content = tokens[idx].content;
+		if (content.includes('<!-- docs: start hidden content -->'))
+			return '<div style="display: none;">';
+		else if (content.includes('<!-- docs: end'))
+			return '</div>';
+		else if (content.includes('<!-- docs: start')) {
```
margaree, line 77: e.g., if `<!-- docs: start accessibility -->` the class would be `d2l-cc-accessibility`

.eleventy.js
margaree: I'll add some checks for length post split just in case the input isn't what we're expecting.

GeorgePlukov (approved):

.eleventy.js
```diff
@@ -1,20 +1,95 @@
...
+			if (content.includes('<!-- docs: live demo -->')) return `${script}<d2l-component-catalog-interactive-demo>${escapeHtml(content)}</d2l-component-catalog-interactive-demo>`;
+			else return `${script}<d2l-component-catalog-demo-snippet-wrapper>${escapeHtml(content)}</d2l-component-catalog-demo-snippet-wrapper>`;
+		} else return defaultFenceRule(tokens, idx, options, env, slf);
+	};
+
+	const defaultTextRule = markdownIt.renderer.rules.text;
+	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
+		const content = tokens[idx].content;
+		if (content.includes('<!-- docs: start hidden content -->'))
+			return '<div style="display: none;">';
```
dlockhart, line 82: I guess there isn't an easy way using this mechanism to just not render the hidden content at all?

.eleventy.js
```diff
@@ -1,20 +1,95 @@
 /* global require, module */
 const cleanCSS = require('clean-css');
 const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
+const { escapeHtml } = require('markdown-it/lib/common/utils');
+
+function _getScript(content) {
```
dlockhart, line 6: Seems a little fragile to parse HTML/JS this way, but probably OK for now. Not sure if there's anything from the [dev-sever](https://github.com/modernweb-dev/web/blob/master/packages/dev-server-core/src/plugins/transformModuleImportsPlugin.ts) that we could use (they have a "es-module-lexer" that looks promising). 

As we discussed, if we just used `dev-server` the imports from the demos would "just work". The more workarounds like this we end up, the more sense it may make to use it directly. 🤷

.eleventy.js
```diff
@@ -1,20 +1,95 @@
...
+	if (scriptStart.length !== 2) return '';
+
+	const scriptContent = scriptStart[0].split('<script type="module">\n');
+	if (scriptContent.length !== 2) return '';
+
+	const importsArray = scriptContent[1].split('\n');
+	let imports = '';
+	importsArray.forEach((importUrl) => {
+		if (importUrl.includes('import ')) {
+			const importStatement = importUrl.replace(/^(.*import.*[\\/])/g, 'import \'/assets/js/'); // replace everything before final backslash
```
dlockhart, line 19: If there's a "view source" option on the demo snippet, would this mean that they'd see the transformed import paths or the original ones?

.eleventy.js
margaree: This was the simplest way to do it - otherwise you have to check each type of markdown content (e.g., text, link, table, etc. - I haven't actually been able to find an exact list of all the token types) and then if some bool hidden is true then you could not render it. If we were going to have pieces that are actual secrets then it could be worth it but so far that seems unlikely.

.eleventy.js
margaree: The source shows the original ones.

.eleventy.js
margaree: Yeah for sure - I'll try out dev-server and see if it helps.

.eleventy.js
dlockhart: Cool yeah seems good for now!

dlockhart (approved):

margaree:
I'll merge this and work on `dev-server` investigation separately so it doesn't make this PR too big/cluttered.

PR #96, opened 2021-04-14
Initial issue template commits

GeorgePlukov:


.github/ISSUE_TEMPLATE/component-in-progress.md
```diff
@@ -0,0 +1,28 @@
+---
+name: Component in progress
+about: If the component is actively being developed or designed use this template
+title: ''
+labels: ''
+assignees: ''
+
```
margaree, line 7: nit: I don't think this extra line is needed

.github/ISSUE_TEMPLATE/documented-component-template.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Documented component template
+about: This template should be used for components that are already designed or developed
+  and will soon be published to the component catalogue
```
margaree, line 4: Will this being on a different line cause any problems?

.github/ISSUE_TEMPLATE/component-in-progress.md
```diff
@@ -0,0 +1,28 @@
+---
+name: Component in progress
```
margaree, line 2: We probably want to be consistent with capitalization for names - e.g., this one should be `Component In Progress`

.github/ISSUE_TEMPLATE/documented-component-template.md
```diff
@@ -0,0 +1,27 @@
+---
+name: Documented component template
```
margaree, line 2: Since we don't include "template" in the other names, maybe something like just `Documented Component` (or we could start them all with "Component" and then have `Component - Request`, `Component - In Progress` and `Component - Documented` so they'd all be grouped together in case we end up creating other issue templates)

.github/ISSUE_TEMPLATE/component-request.md
```diff
@@ -0,0 +1,43 @@
+---
+name: Component Request
+about: Submit a request for a new component to be built
+title: ''
+labels: Component Request
```
margaree, line 5: I noticed only this template would be getting a label. Since we're using the idea of a published label once a component has been tested in dev, we could add the appropriate labels to all 3 so it's easy for us to tell what status someone thinks a component is at a glance after they create the issue.

margaree (commented):
These are looking good! The content seems useful.

.github/ISSUE_TEMPLATE/documented-component-template.md
margaree: I think the text should be more of "already designed and mostly or fully developed" since they'll need to be mostly complete to get a complete page with demos and such in the catalog.

.github/ISSUE_TEMPLATE/documented-component-template.md
GeorgePlukov: Yeah I like having the `Component -` as the prefix, having them grouped together if we add more is a plus

margaree (approved):

PR #97, opened 2021-04-15
Use web-dev-server

margaree:
I used this blog post https://griffa.dev/posts/using-web-components-with-11ty/ quite a bit for piecing this together. 

dev works fine, and I think would also work if we removed the `getPath` piece which would simplify things further.
prod needs `nodeResolve` to be true (not sure of the impact with this?), or would also work if I add back `_getScript` to `.eleventy.js` but replace with `getPath` rather than `/assets/js`.

If having `nodeResolve` be true is fine then I think this can get simpler, otherwise I'll add in the script portion.


dlockhart (approved):
Cool! So best of both worlds? We can use bare imports everywhere, dev server in development and then no need for it in production? Sounds amazing.

margaree:
Mhm! It's pretty nice. So I think then that I should add in the script to replace the imports for demos in prod (since otherwise those wont work as expected unless we're using `nodeResolve` which I don't think we want).

package.json
```diff
@@ -2,15 +2,14 @@
...
-    "build:rollup": "rimraf assets/js && rollup -c rollup.config.js",
-    "clean": "rimraf _site",
+    "build": "rimraf _site && npm run build:rollup && NODE_ENV=production npx eleventy",
+    "build:rollup": "rollup -c rollup.config.js",
     "lint": "npm run lint:eslint",
     "lint:eslint": "eslint . --ext .js",
-    "serve": "npm-run-all clean build:rollup serve:eleventy",
-    "serve:eleventy": "npm run clean && cross-env ELEVENTY_ENV=development eleventy --serve",
-    "test": "npm run lint"
+    "start": "rimraf _site && npm run build:rollup && eleventy && concurrently \"npx eleventy --watch\" \"web-dev-server --node-resolve --open --watch\"",
```
margaree, line 9: There's an extra `eleventy` here in order to build the site so that when dev-server opens the page it isn't `Not Found` because it opens the page before the site finished building (this isn't really necessary but I thought it would avoid confusion and it doesn't really slow anything down)

PR #99, opened 2021-04-20
Parse GitHub issues to generate component doc pages

margaree:
This PR adds the functionality to parse a component GitHub issue and then generate necessary information from it. For example:
- Install dependency/dependencies
- Copy custom-elements.json info
- Add info to file for generating status table
- Copy markdown file to pages directory and add front matter to top
- Copy screenshots
- If `NODE_ENV` is production then only include components with "Published" label

 The main change is the new generate-pages.js script. There are also some other related tweaks and some semi-unrelated style tweaks.


tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+		let showComponent = process.env.NODE_ENV !== 'production'; // if NODE_ENV is production, only show component with "Published" label
+		issue.labels.forEach((label) => {
+			if (label.name === name) labelMatch = true;
+			if (label.name === ISSUE_LABELS.PUBLISHED) showComponent = true;
+		});
+		return issue.labels && issue.labels.length > 0 && labelMatch && showComponent;
+	});
+}
+
+function _getMissingDependenciesList(dependencies) {
```
margaree, line 93: Installing dependencies was a bit annoying because of the async-ness of it. In order to not install over an existing install I try to list the dependency, then if it doesn't exist it gets installed.

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+	_getMissingDependenciesList(dependencies).then((res) => {
+		exec(`npm i ${res} --no-save`, (err, stdout) => {
+			if (err) console.log(`ERROR: Failed to install dependency: ${err.message}`);
+			console.log(`INFO: installation stdout: ${stdout}`);
+			callback();
+		});
+	});
+}
+
+function _parseComponentIssueInfo(componentBody, name, url) {
```
margaree, line 122: This will mostly be used for the status table.

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
+const { exec } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const request = require('request');
+
+const DIR_GENERATED = path.join(__dirname, '../.generated');
```
margaree, line 6: I just put anything that might change in a variable at the top so it's easy to find.

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+	});
+
+	inProgress.forEach((issue) => {
+		const body = _splitIssueInfo(issue.body);
+		componentIssues.push(_parseComponentIssueInfo(body, issue.title, issue.html_url));
+	});
+
+	documented.forEach((issue) => {
+		try {
+			const issueInfo = _splitIssueInfo(issue.body).split(/\r?\n\r?\n/);
```
margaree, line 182: Splits issue comment section based on empty line where front matter is above and other information is below (slightly fragile).

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+			const frontMatter = issueInfo[0];
+			const issueBody = issueInfo[1];
+			const info = _parseComponentIssueInfo(issueBody, issue.title, issue.html_url);
+
+			info.components = JSON.parse(info.components);
```
GeorgePlukov, line 187: Might be good to add a null check here for `info.components`  in case the section is missing. edit: maybe we should do this for all of the `info.****` items

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+				rollupFiles.push(`${info.baseInstallLocation}/${component}`);
+			});
+
+			const dirname = path.dirname(`${info.baseInstallLocation}/${info.markdown}`);
+			const screenshotPath1 = path.join(dirname, 'screenshots');
+			const screenshotPath2 = path.join(dirname, '../screenshots'); // some repos have screenshots in a sibling directory to docs
+			if (!screenshotLocations.includes(screenshotPath1)) screenshotLocations.push(screenshotPath1);
+			if (!screenshotLocations.includes(screenshotPath2)) screenshotLocations.push(screenshotPath2);
+
+			markdownFiles.push({ name: issue.title, file: `${info.baseInstallLocation}/${info.markdown}`, frontMatter: frontMatter });
```
GeorgePlukov, line 198: Maybe we should use the eleventyNavigation -> title here to ensure the title in the navigation page matches exactly to the component title (instead of trusting that the issue title matches identically).

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+const STATES_COMPONENT = {
+	NOT_STARTED: 'Not Started'
+};
+
+function _copyCustomElements(repos) {
+	let tags = [];
+	repos.forEach((repo) => {
+		const customElementsFilePath = path.join(__dirname, `../node_modules/${repo}/custom-elements.json`);
+		if (!fs.existsSync(customElementsFilePath)) {
+			console.log(`WARNING: custom-elements.json does not exist for ${repo}`);
```
GeorgePlukov, line 34: nit: Since we are using `console.error` we might want to use `console.warn` and `console.info` for other logging.

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+		const newFile = path.join(DIR_IMPORTED_COMPONENTS, `${file.name}.md`);
+		const content = fs.readFileSync(originFile);
+		fs.writeFileSync(newFile, `${file.frontMatter}\n${content}`);
+	});
+}
+
+function _copyScreenshots(screenshotDirs) {
+	screenshotDirs.forEach((dir) => {
+		const fullDirectory = path.join(__dirname, `../node_modules/${dir}`);
+		if (!fs.existsSync(fullDirectory)) return;
```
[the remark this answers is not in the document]
GeorgePlukov: If they are always in one of these two spots I don't think its too bad to just check for either of them. Adding another piece of configuration seems a bit unnecessary

package.json
```diff
@@ -2,12 +2,14 @@
   "name": "d2l-component-catalog",
   "description": "",
   "scripts": {
-    "build": "rimraf _site && npm run build:rollup && NODE_ENV=production npx eleventy",
+    "build:dev": "npm run clean && NODE_ENV=development node ./tools/generate-pages.js",
```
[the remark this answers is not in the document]
margaree: Okay cool, I'll try that!

tools/generate-pages.js
margaree: So the eleventy title and the issue title don't actually need to match. The title used in navigation will just be whatever is in eleventyNavigation and the file name will basically just be ignored. Something I have run into though is that with the date/time input components I want to call the issue something like "Input Date & Time Components" but then that ends up as a weird filename, so I think using the eleventy navigation key for filenames would make sense.

.eleventy.js
```diff
@@ -31,6 +31,14 @@ module.exports = function(eleventyConfig) {
 				case 'link_close':
 					token.tag = 'd2l-link';
 					break;
+				case 'heading_open': {
+					let headingClass = token.attrGet('class') || '';
+					if (token.tag === 'h1') headingClass += 'd2l-heading-1';
```
dlockhart, line 36: Won't this be missing a space between the initial value from `token.attrGet('class')` and `d2l-heading-1`?

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
+const { exec } = require('child_process');
+const fs = require('fs');
+const path = require('path');
+const request = require('request');
```
dlockhart, line 4: Curious if we can use ES module imports for these scripts now that Node.js supports them? Or do the individual libraries need to opt into that...

tools/generate-pages.js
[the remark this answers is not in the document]
dlockhart: And I guess depending on where we land with interactive demos, we may be able to ignore screenshots in the future.

tools/generate-pages.js
```diff
@@ -0,0 +1,231 @@
...
+		exec(`npm i ${res} --no-save`, (err, stdout) => {
+			if (err) console.log(`ERROR: Failed to install dependency: ${err.message}`);
+			console.log(`INFO: installation stdout: ${stdout}`);
+			callback();
+		});
+	});
+}
+
+function _parseComponentIssueInfo(componentBody, name, url) {
+	const lines = componentBody.split(/\r?\n/);
```
dlockhart, line 123: Wondering if there might be a front-matter parser [similar to this one](https://www.npmjs.com/package/parser-front-matter) that could do the trick here? Just to avoid all the splitting, etc.

tools/generate-pages.js
margaree: This does work if I change the file extension to `.mjs`. Is that expected or is there a step I missed?

package.json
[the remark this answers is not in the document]
margaree: Just adding `cross-env NODE_ENV=production` before `npm run build` in a script does actually seem to work, which I really thought I'd tried before, so I think I'll just leave it like that since I don't think this usage needs anything more complicated.

.github/ISSUE_TEMPLATE/component-documented.md
```diff
@@ -7,9 +7,22 @@ assignees: ''
...
+keywords:
+  - <possible search terms>
+eleventyNavigation:
+  key: <component name, lower case and hyphenated (e.g., menu-item)>
+  title: <component name, capitalized (e.g., Menu Item)>
+  parent: <component type>
+---
+
+repo: "<url>"
+baseInstallLocation: "<npm install location (e.g., @brightspace-ui/core)>"
```
margaree, line 21: In order to use `gray-matter` these need to be wrapped in quotes, as if there are certain characters (e.g., `@`) the parser fails.

tools/generate-pages.mjs
```diff
@@ -0,0 +1,241 @@
...
+		exec(`npm i ${res} --no-save`, (err, stdout) => {
+			if (err) console.error(`ERROR: Failed to install dependency: ${err.message}`);
+			console.info(`INFO: installation stdout: ${stdout}`);
+			callback();
+		});
+	});
+}
+
+function _parseComponentIssueInfo(componentBody, name, url) {
+	const info = matter(`---\n${componentBody}\n---`).data;
```
margaree, line 133: This is mildly hacky - I wrapped the component body (the info that's below the front matter in the comment) in the front matter delimiter so that matter would parse it out into an object. We could also require that each piece of information in the issue be wrapped in `"` then wrap the body string in `{}` and the we could do a `JSON.parse`. All the quotes would be a bit cluttered but it should work too.

package.json
```diff
@@ -33,6 +33,7 @@
     "eslint-plugin-html": "^6",
     "eslint-plugin-lit": "^1",
     "eslint-plugin-sort-class-members": "^1",
+    "gray-matter": "^4",
```
GeorgePlukov, line 36: We should add request as a devDependency I think

GeorgePlukov (approved):

package.json
margaree: I swear I had that at some point and then must have accidentally deleted it🤦‍♀️

PR #102, opened 2021-04-28
Example Component Page - Part 1

margaree:
This PR begins the process of styling a component page like the design. There is still some work to do such as some padding fixes, navigation accordion, adding tag to heading, best practices icons, but I thought I'd split it out just so the PRs don't get too big and convoluted. There are some miscellaneous tweaks made as well (I'll add some comments throughout).

I created component issues in documentation for date & time components and for button components. I have a branch in core  (https://github.com/Brightspaceui/core/tree/input-date-time-catalog) which adds the corresponding markdown files. Button is more correct to the design.

I'll work on a part 2 to this as well that's making the other changes mentioned.

Also currently best practices are living in the Button md file but I think those seem like they should live in the `design` directory and be pieced together since I think they are more likely to be changed by a designer.


.eleventy.js
```diff
@@ -6,6 +6,11 @@ const { getScript } = require('./tools/getScript');
 
 module.exports = function(eleventyConfig) {
```
margaree, line 7: I added a favicon partially just to see how to do it and partially so it looked nicer in the tabs. Can be changed later (just used the D2L one).

The "enhancements" shortCode is used to add the "Looking for an enhancement..." text to all the component markdown page. This could also be something that is in the actual markdown document instead, but I thought this was a bit better in case we want to change the messaging.

package.json
```diff
@@ -2,8 +2,8 @@
   "name": "d2l-component-catalog",
   "description": "",
   "scripts": {
-    "build": "npm run clean && node ./tools/generate-pages.mjs",
-    "build:prod": "cross-env NODE_ENV=production npm run build && npm run build:rollup && cross-env NODE_ENV=production npx eleventy",
+    "build": "npm run clean && node ./tools/generate-pages.mjs && npm run build:rollup",
```
margaree, line 5: `build:rollup` was moved to be in `build` so that the manifest would be generated in dev as well so that someone wouldn't think everything was fine in dev but then in prod a file wasn't actually being included (since in dev an import can just be retrieved from node_modules without having needed to be rolled up).

pages/_data/assets.js
```diff
@@ -1,15 +1,11 @@
...
-			const assets = require('../_includes/manifest.json');
-			const modulePath = assets[assetName];
-			if (!modulePath) {
-				throw new Error(`Error with getPath: ${assetName} does not exist in manifest.json`);
-			}
-			return `/${modulePath}`;
-		} else {
-			return `/pages/assets/${assetName}`;
-		}
+		const assets = require('../_includes/manifest.json');
```
margaree, line 4: Moved this because of same situation mentioned above.

pages/_includes/layouts/base.njk
```diff
@@ -3,45 +3,74 @@
 	{% include "css/prism-ghcolors.css" %}
 {% endset %}
 <!doctype html>
-<html lang="en">
+<html class="initial-hide" lang="en">
```
margaree, line 6: This is to get rid of the FOUC.

pages/_includes/layouts/base.njk
```diff
@@ -3,45 +3,74 @@
...
 			<nav aria-label="Component Information">
 				{% set navPages = collections.all | eleventyNavigation %}
-				{% macro renderNavListItem(entry) -%}
-					<li>
-						<d2l-link href="{{ entry.url | url }}" {% if entry.url == page.url %} class="d2l-component-catalog-link-selected"{% endif %}>{{ entry.title }}</d2l-link>
+				{% macro renderNavListItem(entry, size) -%}
+					<li {% if entry.url == page.url %} class='d2l-component-catalog-li-selected' {% endif %}>
+						<a
+							href="{{ entry.url | url }}"
+							{% if entry.url == page.url %}
```
margaree, line 34: This is all a bit verbose - I'm not overly familiar with nunjucks but I'll look into if I can simplify this.

tools/generate-pages.mjs
```diff
@@ -217,14 +214,23 @@ request(ISSUES_REQUEST, (error, response, body) => {
 
 	console.info('INFO: Completed GitHub issue processing');
 
+	fs.mkdirSync(DIR_GENERATED, { recursive: true });
```
margaree, line 217: This is just moved around so that if there aren't dependencies to install (i.e., in prod if there is nothing documented with the 'Published' label) the build wont fail on the rollup step and can still do the steps that don't involve dependencies.

.eleventy.js
dlockhart: Yeah I'm definitely excited to remove that blurb from all our Markdown files and centralize it somewhere in here. I could see us changing this up a bit, but this is good for now!

pages/_includes/layouts/base.njk
dlockhart: `display:none` will prevent all layout and paints from happening, which is maybe fine. In the monolith we use an `opacity: 0;` on the body with a transition that fades it in when it's removed.

pages/_includes/layouts/base.njk
```diff
@@ -3,45 +3,74 @@
...
-		<div>
-			Brightspace Component Catalog
-		</div>
-	</header>
 	<main>
-		<div class="d2l-component-catalog-side-nav d2l-body-compact">
+		<div class="d2l-component-catalog-side-nav d2l-body-standard">
+			<div class="d2l-component-catalog-band"></div>
+			<div class="d2l-component-catalog-title">
+				<img src="/img/logo_slim.png"></img>
```
dlockhart, line 26: Images don't need a closing tag, and this should have an `alt`.

pages/assets/base-imports.js
```diff
@@ -1,2 +1,12 @@
-import '@brightspace-ui/core/components/link/link.js';
+import { linkStyles } from '@brightspace-ui/core/components/link/link.js';
+import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/typography/typography.js';
+
+const styleSheet = document.createElement('style');
+styleSheet.type = 'text/css';
+styleSheet.innerText = linkStyles;
+document.head.appendChild(styleSheet);
```
dlockhart, line 8: Would creating some web components for the site structure & navigation avoid needing to do create this dynamic stylesheet?

margaree: Yeah I think especially a component for navigation would be helpful. I've avoided it a bit because I'm unsure how it would work with `eleventyNavigation` but I think this could overall be helpful so I'll give that a go.

pages/assets/navigation.js
```diff
@@ -0,0 +1,91 @@
...
+			}
+			ul ul li {
+				padding-top: 0;
+			}
+		`];
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+		this._navItems = JSON.parse(this.navItems);
```
GeorgePlukov, line 52: Maybe we can parse this outside of the component and just pass the array of items through?

GeorgePlukov (approved):
Left a comment but LGTM

pages/assets/navigation.js
margaree: So I think (as in I wasn't able to figure out another way - could totally be missing something) that in order to pass navItems in here https://github.com/BrightspaceUI/documentation/pull/102/files#diff-bf6fd62344d84804cccf1706b1ed0fc72377913b9ce294f4e4603a471160581cR24 that it does need to be a string, which is a bit gross.

margaree: I haven't been able to simplify this further. I'm going to merge this and revisit later on when I'm hopefully more familiar with nunjucks.

PR #104, opened 2021-05-06
Component Page: Part 2

margaree:
The corresponding branch in core is here https://github.com/BrightspaceUI/core/tree/input-date-time-catalog

Notable changes:
- I added a `design` folder to core which would contain a design document for each component. Then a designer can be code owner for those files
  - Current strategy for using those is adding `designMarkdown` to the commented area of the issue, and then that gets pulled into the readme markdown file at the location of the `<!-- docs: design -->` comment. This could change if there is other design information that should appear at different spots in the file, but this seemed like a simple place to start.
- Code blocks are now using the `d2l-code-view` component since we want the styles shared there
- Component tags appear next to the component heading. These come from the document front matter (since just having them in the readme makes it difficult to style them differently than the heading)


.eleventy.js
```diff
@@ -58,14 +58,13 @@ module.exports = function(eleventyConfig) {
...
-	const defaultFenceRule = markdownIt.renderer.rules.fence;
-	markdownIt.renderer.rules.fence = (tokens, idx, options, env, slf) => {
+	markdownIt.renderer.rules.fence = (tokens, idx) => {
 		const content = tokens[idx].content;
 		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->')) {
 			const script = getScript(content, process.env.NODE_ENV);
 			if (content.includes('<!-- docs: live demo -->')) return `${script}<d2l-component-catalog-interactive-demo>${escapeHtml(content)}</d2l-component-catalog-interactive-demo>`;
 			else return `${script}<d2l-component-catalog-demo-snippet-wrapper>${escapeHtml(content)}</d2l-component-catalog-demo-snippet-wrapper>`;
-		} else return defaultFenceRule(tokens, idx, options, env, slf);
+		} else return `<d2l-component-catalog-code-view-wrapper>${escapeHtml(content)}</d2l-component-catalog-code-view-wrapper>`;
```
margaree, line 67: Uses a wrapper component in order to use `d2l-code-view`

.eleventy.js
```diff
@@ -80,6 +79,15 @@ module.exports = function(eleventyConfig) {
 			const splitEnd = splitStart[1].split(' -->');
 			const contentClass = `d2l-component-catalog-${splitEnd[0].replace(/ /g, '-')}`;
 			return `<div class="${contentClass}">`;
+		} else if (content.includes(':x:')) {
```
margaree, line 82: In this file https://github.com/BrightspaceUI/core/blob/input-date-time-catalog/design/buttons.md I wanted to use icons so that there was a way of quickly visually indicating the dos and don't without having the words "Do" and "Don't" above the lists, since those aren't in the design, just in case someone actually looks at this file in GitHub rather than in the component catalog. We don't actually want to use these icons though in the file so this removes them. I don't love this and am open to alternatives here.

pages/_includes/css/style.css
```diff
@@ -3,18 +3,20 @@ html {
...
-	transition: opacity ease-in 0.2s;
-}
-body.d2l-component-catalog-body-unresolved {
-	opacity: 0;
 }
 main {
 	display: flex;
 	min-height: 100vh;
 }
+div.d2l-component-catalog-content.d2l-component-catalog-content-unresolved {
```
margaree, line 11: I moved the unresolved piece to be just the content rather than the body. This prevents the flicker in the navigation side bar when changing pages in the prod site.

pages/_includes/css/style.css
```diff
@@ -57,6 +59,21 @@ main {
 .d2l-component-catalog-content img {
 	max-width: 100%;
 }
+.d2l-component-catalog-content p:empty {
```
margaree, line 62: `markdown-it` sometimes adds empty paragraphs so I'm removing the margins that are unnecessary in those cases. I think this is related to the lines that just contain comments.

pages/_includes/css/style.css
```diff
@@ -70,8 +87,28 @@ d2l-component-catalog-interactive-demo {
...
 	margin: 0;
-	padding-left: 1rem;
+}
+.d2l-component-catalog-best-practices li::before {
+	font-weight: bold;
+	display: inline-block;
+	margin-left: -1.3em;
+	width: 1em;
+}
+.d2l-component-catalog-dos li::before {
```
margaree, line 100: I couldn't figure out a simple way to use our d2l-icons as the marker while keeping this as a list. This uses unicode checkmarks and xs, which looks quite similar.

pages/_includes/layouts/base.njk
```diff
@@ -12,18 +11,21 @@
...
-<body class="d2l-typography d2l-component-catalog-body-unresolved">
+<body class="d2l-typography">
 	<main>
-		<div class="d2l-component-catalog-side-nav d2l-body-standard">
+		<div class="d2l-component-catalog-side-panel d2l-body-standard">
 			<div class="d2l-component-catalog-band"></div>
-			<div class="d2l-component-catalog-title">
-				<img src="/img/logo_slim.png" alt="Brightspace Logo" />
-			</div>
+			<a href="/" title="Home">
```
margaree, line 20: I made the logo navigate to the home page since that seems to be common practice.

pages/assets/base-imports.js
```diff
@@ -1,8 +1,8 @@
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/link/link.js';
 import '@brightspace-ui/core/components/typography/typography.js';
-import './navigation.js';
 
-document.fonts.ready.then(() => {
-	document.body.classList.remove('d2l-component-catalog-body-unresolved'); // prevent FOUC
+window.addEventListener('DOMContentLoaded', () => {
```
margaree, line 5: I found that listening for `DOMContentLoaded` more ensure that the user was not going to see a flash of the unstyled demo snippets.

pages/assets/code-view-wrapper.js
```diff
@@ -0,0 +1,34 @@
...
+			}
+			:host([hidden]) {
+				display: none;
+			}
+		`;
+	}
+
+	render() {
+		return html`
+			<d2l-code-view language="html" hide-language><div class="d2l-component-catalog-code-view-wrapper-content"><slot @slotchange="${this._getCode}"></slot></div>
```
margaree, line 19: This seemed to need to be on one line, otherwise there was an unnecessary indentation in the code block.

pages/assets/navigation.js
```diff
@@ -21,29 +22,50 @@ class ComponentCatalogNavigation extends LitElement {
 			:host([hidden]) {
 				display: none;
 			}
+			.d2l-component-catalog-hidden {
```
margaree, line 24: Changes in this file are mostly tweaks to get the navigation sidebar looking like the design.

tools/generate-pages.mjs
```diff
@@ -46,14 +48,25 @@ function _copyCustomElements(repos) {
 
 function _copyMarkdown(files) {
 	files.forEach((file) => {
-		const originFile = path.join(__dirname, `../node_modules/${file.file}`);
```
margaree, line 49: Changes to merge the design document into the dev document.

.eleventy.js
margaree: I believe so - it represents the case where code isn't a demo or anything and is just a chunk like
```
<my-component>
```
and we want to use our `d2l-code-view` (or whatever that gets replaced with - the wrapper allows for that to be switched out fairly easily; could even probably just use prism on our own and just import the shared styles) so that we get the correct styling.

I don't think there are other cases that we would have in our docs that would trigger a fence where we wouldn't want to use code-view, though I could be missing something.

GeorgePlukov (approved):

.eleventy.js
margaree: I think I'll actually remove the icons and this section. It's fairly doubtful that someone would actually look at the markdown file in core/other repos, and if they do, the wording is still clear on do vs don't.

pages/_includes/layouts/base.njk
```diff
@@ -1,29 +1,31 @@
...
+				<div class="d2l-component-catalog-title">
+					<img src="/img/logo_slim.png" alt="Brightspace Logo" />
```
dlockhart, line 22: Nit: since it's acting as a home link now, the alt should probably read "Home" or something that will better explain what will happen when you click it.

dlockhart (approved):

PR #105, opened 2021-05-13
Add GitHub action that will run when Published label added/removed

margaree:
Objective: Gain more control over a Published page appearing in the component catalog in prod

Details:
- Documented component will only have a page in the prod site if it has the "Published" label AND if its issue number appears in the list in `tools/component-page-issues.js`
- I added a GitHub action that is triggered when a label is added or removed from an issue. If that label is "Published", then the issue number will be added or removed as necessary from the file and a PR will be opened. If there is already an open PR for this then the change will just be pushed and the description will be updated with the new change will be appended to the end

Questions:
- Should the action in `update-labeled-issues-list` live here or in `BrightspaceUI/actions`? I tried to make this semi-generic in case.
- I added myself and George as CODEOWNERS so that we'll just be auto-added as reviewers on these PRs. My thought is that eventually this should be gaudi-dev once this is more in maintenance mode. Any differing thoughts?


svanherk:
> Should the action in update-labeled-issues-list live here or in BrightspaceUI/actions? I tried to make this semi-generic in case.

I still need to go through the whole PR but I _think_ it makes sense to just live here for now, until we have a use case for it somewhere else.  The solution seems fairly specific for what we need to do here, but if it works really nicely maybe it'll come up in the future for something.  That technically means it could be all in one giant workflow instead of an action, but I think leaving it as an action is fine in case we ever do want to move it.  Plus it gave you some practice with both!

>I added myself and George as CODEOWNERS so that we'll just be auto-added as reviewers on these PRs. My thought is that eventually this should be gaudi-dev once this is more in maintenance mode. Any differing thoughts?

That makes sense to me!

GeorgePlukov (approved):
> I added myself and George as CODEOWNERS so that we'll just be auto-added as reviewers on these PRs. My thought is that eventually this should be gaudi-dev once this is more in maintenance mode. Any differing thoughts?
Sounds good to me

I haven't used github actions much in the past but this seems to be doing what we want!

.github/workflows/issue-label-change.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  sync_components:
+    name: Sync Component Pages List
+    timeout-minutes: 5
+    runs-on: ubuntu-latest
+    steps:
+      - uses: Brightspace/third-party-actions@actions/checkout
+      - uses: Brightspace/third-party-actions@actions/setup-node
+      - name: Update Published Components File
+        uses: ./.github/workflows/update-labeled-issues-list/
```
svanherk, line 14: I wonder if this has to be here, or if we could move it to `./.github/actions/update-labeled-issues-list/` to keep the action separate from the workflow.  I _think_ it's only workflows that need to exist in a certain place

.github/workflows/update-labeled-issues-list/action.yml
```diff
@@ -0,0 +1,86 @@
...
+    required: true
+  LIST_PATH:
+    description: Path to file containing list of issue numbers that have LABEL_NAME label on them
+    required: true
+runs:
+  using: composite
+  steps:
+      - name: Install Dependencies
+        run: |
+          echo -e "\e[34mInstalling Dependencies"
```
svanherk, line 18: Hehe yeah I can tell you used `visual-diff` as an example 😛 .  This `\e##m` stuff is so ugly but the output looks so nice!

.github/workflows/update-labeled-issues-list/action.yml
```diff
@@ -0,0 +1,86 @@
...
+          npm install chalk@4 @octokit/rest@18 --prefix ${{ github.action_path }} --no-save --loglevel error
+        shell: bash
+      - name: Check out branch
+        id: check-out-branch
+        run: |
+          BRANCH_NAME="update-issue-list"
```
svanherk, line 24: Super Minor: In `BSI` and `visual-diff` the reason I had `ghworkflow/` as part of the branch name was to pick something really specific no one would ever accidentally use, and then have their branch overwritten.  I think the odds are super low regardless, considering we shouldn't be updating the list manually

.github/workflows/update-labeled-issues-list/action.yml
```diff
@@ -0,0 +1,86 @@
...
+          git fetch origin +refs/heads/$BRANCH_NAME:refs/heads/$BRANCH_NAME || true
+
+          echo -e "\e[34mChecking out branch"
+          git checkout $BRANCH_NAME || git checkout -b $BRANCH_NAME
+        shell: bash
+      - name: Update issue list
+        run: |
+          if [ ${{ github.event.label.name }} != "${{ inputs.LABEL_NAME }}" ]; then
+            echo -e "\e[32mLabel changed is not \"${{ inputs.LABEL_NAME }}\""
+            exit 0;
```
svanherk, line 37: So this is what I was trying to remember in the demo but then couldn't... I think this just exits and goes to the next step, but doesn't exit the whole run correct?  Looks like that's ok, it will just say "No changes to commit" in the "Committing changes" step and "No changes to PR" in the "Opening PR" step. But you could also set an output here and check that as the first step of every remaining step if you want no further output.

.github/workflows/update-labeled-issues-list/handle-pr.js
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env node
```
svanherk, line 1: As also mentioned in the demo, there's lots of helper actions that you could use to avoid these JavaScript files if we were to just do this all in a workflow, which would look more similar to the [BSI Update PR workflow](https://github.com/Brightspace/brightspace-integration/blob/master/.github/workflows/bsi-update.yml#L57-L89).  But that's to make things easier really - if you've already done the "hard part" I don't think there's a problem with leaving this as an action for future flexibility/reuse.  (For example - we've played around with the idea of rewriting the BSI one as an action to help keep any repo's `package-lock.json` up-to-date)

.github/workflows/update-labeled-issues-list/handle-pr.js
```diff
@@ -0,0 +1,68 @@
...
+		}
+	} else {
+		console.log(`PR for branch ${branchName} already exists. Appending additional information to body.`);
+		const prNum = existingPR.data[0].number;
+		try {
+			await octokit.request('PATCH /repos/{owner}/{repo}/pulls/{pull_number}', {
+				owner: owner,
+				repo: repo,
+				pull_number: prNum,
+				body: getPRBody(existingPR.data[0].body)
```
svanherk, line 56: Probably an edge case, but if someone was toggling a label could this create some weird PR body that goes like:
```
Summary of changes. Please review impact of changes before merging
Added: Button
Removed: Button
Added: Button
```
(Not sure what issue title looks like exactly)

Probably not worth changing - this might even be a feature lol, if we want to know this is happening and look into why.

.github/workflows/issue-label-change.yml
margaree: That sounds like it would work - I'll test it out

.github/workflows/update-labeled-issues-list/action.yml
margaree: Ah good point

.github/workflows/update-labeled-issues-list/action.yml
margaree: Yeah so it does keep going to the next steps - a way to just completely exit early would be really nice. The output might be the cleaner approach just so it isn't running things unnecessarily. The other thing it sounded like I could do based on your demo would be to add an `if` condition to the workflow (though if we wanted this to be used other places then that might not work, if they used different labels, but at the moment that seems unlikely).

.github/workflows/update-labeled-issues-list/handle-pr.js
margaree: Yup that definitely could happen. I had considered parsing the body for the patch changes and then just having one list for added and one for removed, but I kind of like that it tells a story and I think it would overall be rare to have a lot of toggling. It also seemed cleaner to not do too much string parsing.

.github/workflows/update-labeled-issues-list/action.yml
svanherk: Yeah good point. If we know we will only care about "published", an `if` at the workflow level seems like a good idea even if the action somewhat handles it for us.  That way we cancel out asap, without checking out the repo, spending time installing dependencies, etc.

svanherk (approved):
Looks good to me!  If we do come up with a future use case we'll probably need to tweak the inputs and add back the label, but for now I think the early cancelling is the way to go! And if that happens we'll have a much better idea of what's needed as inputs anyways.

PR #139, opened 2021-05-27
Changes related to adding core components

margaree:
These were tweaks I found were either useful or needed to get the core components working once their pages were added. These don't rely on any core changes and look fine with core on master branch so this can merge whenever.

General Changes:
- Get table working (added table component in order to use wrapper and styles) [let me know if there are thoughts around a better way to do this]
- Remove the design markdown joining piece (as discussed in core PR) - can be re-added later on if found to be useful
- Use Octokit in order to fetch issues (paging)
- Add component type pages


pages/assets/navigation.js
```diff
@@ -84,14 +84,22 @@ class ComponentCatalogNavigation extends LitElement {
 		`;
 	}
 
-	_generateList(entry, size) {
+	_generateList(entry, size, isComponentsGrandchildren) {
```
margaree, line 87: I alphabetized the components within a component type since that looks cleaner.

pages/components/actions.md
```diff
@@ -4,6 +4,7 @@ eleventyNavigation:
   title: Actions
   key: actions
   parent: components
+  order: 2
```
margaree, line 7: I specified the order for component type pages based on Jeff's design. `order: 1` is for the Status page.

tools/generate-pages.mjs
```diff
@@ -84,17 +72,6 @@ function _copyScreenshots(screenshotDirs) {
 	});
 }
 
-function _generateRollupConfig(files) {
```
margaree, line 87: Cleanup - didn't actually need this, could just use `_writeJSONToGeneratedFile` function

tools/generate-pages.mjs
```diff
@@ -105,14 +82,12 @@ function _getCommentContent(body) {
 function _getIssues(issues, expectedLabelName) {
 	const isProd = process.env.NODE_ENV === 'production';
 	return issues.filter(issue => {
+		if (isProd && !publishedComponents.includes(issue.number)) return false;
```
margaree, line 85: Checking for the `Published` label was moved to when issues are being requested so that allowed this to be cleaned up a bit.

pages/assets/table.js
```diff
@@ -0,0 +1,39 @@
+import { css, html, LitElement } from 'lit-element';
+import { tableStyles } from '@brightspace-ui/core/components/table/table-wrapper.js';
```
dlockhart, line 2: I ran into a similar situation in the monolith where it was rendering out `<d2l-table-wrapper><table class="d2l-table>` things but I couldn't easily wrap them in a Lit component to include the styles.

I ended up importing the styles separately and adding them to the document, like this:
https://github.com/Brightspace/brightspace-integration/blob/master/web-components/d2l-table.js

A similar approach might work here. You could also turn the table styles into CSS that gets included as part of the app's main CSS.

margaree: The importing styles separately strategy seems to work nicely here. I'll update that.

tools/generate-pages.mjs
```diff
@@ -150,6 +125,21 @@ function _parseComponentIssueInfo(componentBody, name, url) {
 	return info;
 }
 
+async function _requestIssues() {
+	const props = {
+		owner: 'BrightspaceUI',
+		repo: 'documentation',
+		state: 'all'
+	};
+	if (process.env.NODE_ENV === 'production') props.labels = ISSUE_LABELS.PUBLISHED;
```
GeorgePlukov, line 134: This is convenient

tools/generate-pages.mjs
```diff
@@ -150,6 +125,21 @@ function _parseComponentIssueInfo(componentBody, name, url) {
...
+
+	let issueData = [];
+	for await (const response of octokit.paginate.iterator(octokit.rest.issues.listForRepo, props)) {
```
GeorgePlukov, line 137: Just wondering do we need to write this as an async iterator? I was thinking since we are just appending the issues data we could do:
`const issueData = await octokit.paginate(octokit.rest.issues.listForRepo, props)`

Not 100% sure that will work but would be a bit simpler if it does

margaree: Yup that works! Definitely simplifies things.

GeorgePlukov (approved):

PR #140, opened 2021-05-27
Resizable preview component and read only code demos

GeorgePlukov:
Adds the new Playground component demo/code view items and removes the old component wrappers. I'll probably be moving a couple of these components into their own files at some points (like the code-view with the button overlay).


.eleventy.js
```diff
@@ -60,21 +60,29 @@ module.exports = function(eleventyConfig) {
 
 	markdownIt.renderer.rules.fence = (tokens, idx) => {
 		const content = tokens[idx].content;
-		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->')) {
-			const script = getScript(content, process.env.NODE_ENV);
+		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->') || content.includes('<!-- docs: code demo -->')) {
+			const imports = getScript(content);
```
dlockhart, line 64: Wondering if the components themselves could be the ones that strip out the imports/comment lines/etc.? That way consumers (which is only this one place I guess) don't need to.

pages/assets/interactive-demo-wrapper.js
```diff
@@ -0,0 +1,87 @@
...
+				</div>
+				<div class="attribute-table-wrapper">
+				<!-- todo: Add attribute table -->
+				</div>
+			</div>
+		`;
+	}
+
+}
+customElements.define('d2l-component-catalog-code-demo', ComponentCatalogDemoSnippetWrapper);
```
dlockhart, line 87: I think the web component name here is incorrect, shouldn't it be `d2l-component-catalog-interactive-demo`? 

Along those lines though, would it make sense to just combine `d2l-component-catalog-interactive-demo` and `d2l-component-catalog-code-demo` into a single component with an `interactive` attribute? They seem really similar in terms of attributes.

pages/assets/code-demo-wrapper.js
```diff
@@ -0,0 +1,79 @@
...
+import 'playground-elements/playground-code-editor';
+import 'playground-elements/playground-preview';
+import 'playground-elements/playground-project';
+import { css, html, LitElement } from 'lit-element';
+
+const MINIMUM_WIDTH = 300;
+class ComponentCatalogDemoSnippetWrapper extends LitElement {
+	static get properties() {
+		return {
+			code: { type: String, attribute: 'code', reflect: true },
```
dlockhart, line 13: Probably doesn't need to reflect? Also you don't need to set `attribute` if the name matches the property name.

pages/assets/code-demo-wrapper.js
```diff
@@ -0,0 +1,79 @@
...
+				position: absolute;
+				padding-right: 10px;
+				right:0;
+			}
+			
+		`;
+	}
+	constructor() {
+		super();
+		this.showCode = true;
```
dlockhart, line 61: Boolean attributes [should be false-by-default](https://github.com/BrightspaceUI/guide/wiki/LitElement-Best-Practices-&-Gotchas#-do-default-boolean-properties-to-false), otherwise there's no way to set them to false. Suggest maybe flipping this around to `hide-code`?

pages/assets/code-demo-wrapper.js
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	render() {
+		return html`
+			<div class="code-demo-container">
```
dlockhart, line 66: Is this `<div>` needed? Don't see any CSS targeting it and the host element is already block.

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+
+		const onPointerup = () => {
+			bar.releasePointerCapture(pointerId);
+			bar.removeEventListener('pointermove', onPointermove);
+			bar.removeEventListener('pointerup', onPointerup);
+		};
+		bar.addEventListener('pointerup', onPointerup);
+	}
+}
+customElements.define('d2l-resizable-demo', ResizableDemo);
```
dlockhart, line 227: Probably best to prefix this with `d2l-component-catalog-*` just in the unlikely scenario where we want to call something `d2l-resizable-*` in the future.

pages/assets/interactive-demo-wrapper.js
```diff
@@ -0,0 +1,87 @@
...
+	}
+	get getModuleImports() {
+		return this.imports;
+	}
+
+	render() {
+		return html`
+			<div class="code-demo-container">
+				
+				<d2l-resizable-demo .code=${this.getCode} .imports=${this.imports}></d2l-resizable-demo>
```
dlockhart, line 71: Was there a reason to use property binding here instead of an attribute binding? Also shouldn't it be `getCode()`?

pages/assets/interactive-demo-wrapper.js
```diff
@@ -0,0 +1,87 @@
...
+		if (splitItems.length === 2) {
+			// todo insert attributes
+			const withAttributes = `${splitItems[0]} -- insert attributes -- ${splitItems[1]}`;
+			return `${withAttributes}`;
+		}
+
+		return `${this.code}`;
+
+	}
+	get getModuleImports() {
```
dlockhart, line 63: Don't think this is used?

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+	static get properties() {
+		return {
+			// Code for the preview IFrame to display
+			code: { type: String },
+			// Necessary imports for the code running in the IFrame
+			imports: { type: String },
+			// Is the preview resizable
+			isResizable : { type: Boolean },
+			// Is a code editor attached to the preview
+			isAttached : { type: Boolean },
```
dlockhart, line 23: Nit: I might call these `resizable` and `attached`. If you keep these names, they'll need `attribute: 'is-resizable'` to make the binding work.

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+			}
+			.small {
+				height: 200px;
+			}
+			.medium {
+				height: 300px;
+			}
+			.large {
+				height: 400px;
+			}
```
dlockhart, line 52: I'd suggest attribute selectors here and you can avoid computed CSS classes:

```css
:host([size="small"]) {
    height: 200px;
}
```

The size property definition will need `reflect: true` though for this to work.

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+			playground-preview {
+				height: 100%;
+			}
+			.preview-container {
+				width: max(${MINIMUM_WIDTH}px, var(--playground-preview-width, 100%));
```
dlockhart, line 57: You may run into some browser compatibility with the `max()` function in CSS... it's a pretty recent addition.

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+			.content-wrapper { 
+				display: inline;
+			}
+			.empty-space {
+				height: 100%;
+				-webkit-overflow-scrolling: touch;
+				overflow-x: hidden;
+				position: relative;
+			}
+			.d2l-template-primary-secondary-divider-handle-line {
```
dlockhart, line 104: I think this primary/secondary CSS isn't used?

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+							</div>
+						</body>
+					</script>	
+					<script filename="index.js" type="sample/js">
+						${this.imports}
+					</script>
+				</playground-project>
+				<div class=${classMap(previewClasses)}>
+					<playground-preview id="preview" style=${styleMap(previewStyles)} project="p1" filename="index.html"></playground-preview>
+					${this.isResizable ?  html`<div class="slider" style=${styleMap(sliderStyles)} @pointerdown=${this._onResizeBarPointerdown}>
```
dlockhart, line 186: Don't think this handles keyboard support, which is probably fine for a first pass. But also likely wouldn't be too hard to add a tab stop and key handlers for left/right and reuse some of the pointer sizing logic. 🤷

pages/assets/resizable-demo.js
```diff
@@ -0,0 +1,227 @@
...
+import { css, html, LitElement } from 'lit-element';
+import { classMap } from 'lit-html/directives/class-map.js';
+import { styleMap } from 'lit-html/directives/style-map.js';
+
+const SLIDER_WIDTH = 35;
+const MINIMUM_WIDTH = 300;
+class ResizableDemo extends LitElement {
+	static get properties() {
+		return {
+			// Code for the preview IFrame to display
```
margaree, line 16: nit: for code comments, in order to be consistent with `core` etc they should be formatted as blocks, e.g.:
```
/**			 
* Code for the preview IFrame to display
*/
```
and `size` can specify `@type` to show its possible values (e.g., https://github.com/BrightspaceUI/core/blob/master/components/tooltip/tooltip.js#L125)

rollup.config.js
```diff
@@ -5,9 +5,10 @@ import outputManifest from 'rollup-plugin-output-manifest';
 
 const componentFiles = [
 	'./pages/assets/base-imports.js',
-	'./pages/assets/code-view-wrapper.js',
-	'./pages/assets/demo-snippet-wrapper.js',
-	'./pages/assets/navigation.js'
+	'./pages/assets/code-demo-wrapper.js',
+	'./pages/assets/navigation.js',
+	'./pages/assets/resizable-demo.js',
+	'./pages/assets/interactive-demo-wrapper.js'
```
margaree, line 11: nit: alphabetical order of files

.eleventy.js
```diff
@@ -60,21 +60,29 @@ module.exports = function(eleventyConfig) {
...
 			if (content.includes('<!-- docs: live demo -->')) {
 				return `
-					${script}
-					<d2l-component-catalog-interactive-demo>
-						${escapeHtml(content)}
+					<d2l-component-catalog-interactive-demo code="${escapeHtml(content)}" imports="${escapeHtml(imports)}">
```
margaree, line 67: nit: For more consistency, would it make sense to name these components all as `d2l-component-catalog-demo-*` and then start their filenames with `demo-*` for more ease of use?

pages/assets/interactive-demo-wrapper.js
```diff
@@ -0,0 +1,87 @@
...
+			}
+			:host([hidden]) {
+				display: none;
+			}
+
+			playground-code-editor {
+				width: 100%;
+				display: inline-block;
+			}
+
```
margaree, line 31: nit: extra line spacing

pages/assets/resizable-demo.js
GeorgePlukov: Didn't think to check browser compatibility for this, thanks for pointing it out

pages/assets/resizable-demo.js
GeorgePlukov: I'l take a look into adding keyboard support, wasn't thinking of it explicitly to start. As you said doesn't look like it will be particularly difficult.

pages/assets/interactive-demo-wrapper.js
GeorgePlukov: So right now they look pretty similar since I haven't done the attribute parsing yet, but once the attribute table is in the interactive demo component will have quite a bit more code to append the control results to the code snippets in the code-view/preview. I'm not against merging the two components, but it may be that it just complicates things. 

A decent method may be to use the interactive demo component wrapper contain a `d2l-component-catalog-code-demo` and just add the table and manipulate the code string.

pages/_includes/layouts/base.njk
```diff
@@ -12,6 +12,7 @@
 	</style>
 	<script src="{{assets.getPath("base-imports.js")}}" type="module"></script>
 	<script src="{{assets.getPath("navigation.js")}}" type="module"></script>
+	<script src="{{assets.getPath("code-demo-wrapper.js")}}" type="module"></script>
```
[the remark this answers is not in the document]
GeorgePlukov: Accidentally left this in from some test cases, probably don't need it included here unless we are going to be adding demos to non component pages

.eleventy.js
```diff
@@ -54,25 +53,30 @@ module.exports = function(eleventyConfig) {
...
-					</d2l-component-catalog-demo-snippet-wrapper>
+					<d2l-component-catalog-demo-snippet hide-code="true" demo-snippet="${escapeHtml(content)}">
+					</d2l-component-catalog-demo-snippet>
 				`;
 			}
-		} else
-			return `<d2l-component-catalog-code-view-wrapper>${escapeHtml(content)}</d2l-component-catalog-code-view-wrapper>`;
+		} else {
+			// Code only snippets
+			// todo: add a prop to hide the demo and only show code
```
GeorgePlukov, line 77: We could add a prop here to only show the code section or write a separate wrapper component for `playground-code-editor` and extract out the theming css

GeorgePlukov:
Ended up moving the `interactive` into the same `demo-snippet` component and just passing in the base code snippet to do all of the manipulation and splitting of imports within the component

.eleventy.js
```diff
@@ -54,25 +53,30 @@ module.exports = function(eleventyConfig) {
...
-		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->')) {
-			const script = getScript(content, process.env.NODE_ENV);
+		if (content.includes('<!-- docs: live demo -->') || content.includes('<!-- docs: demo -->') || content.includes('<!-- docs: code demo -->')) {
 			if (content.includes('<!-- docs: live demo -->')) {
 				return `
-					${script}
-					<d2l-component-catalog-interactive-demo>
-						${escapeHtml(content)}
+					<d2l-component-catalog-demo-snippet interactive demo-snippet="${escapeHtml(content)}">
 					</d2l-component-catalog-interactive-demo>
```
margaree, line 60: Looks like this closing tag should be `d2l-component-catalog-demo-snippet`

pages/_includes/css/style.css
```diff
@@ -76,7 +76,6 @@ h1, h2, h3 {
 }
 
 /** Specialty Content Sections */
-d2l-component-catalog-demo-snippet-wrapper,
 d2l-component-catalog-interactive-demo {
```
margaree, line 79: I believe this can now be removed as well.

pages/_includes/layouts/component.njk
```diff
@@ -3,8 +3,8 @@ layout: layouts/base
 permalink: components/{{eleventyNavigation.parent}}/{{eleventyNavigation.key}}.html
 ---
 {{ content | safe }}
-<script src="{{assets.getPath("code-view-wrapper.js")}}" type="module"></script>
-<script src="{{assets.getPath("demo-snippet-wrapper.js")}}" type="module"></script>
+<script src="{{assets.getPath("demo-resizable-preview.js")}}" type="module"></script>
```
margaree, line 6: If demo-resizable-preview is only used within demo-snippet then I think you don't need to import it separately.

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			.slider-region {
+				height: 100%;
+				width:100%;
+				z-index:5;
+			}
+			.code-demo-container {
```
margaree, line 52: I don't think this class is used

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
+import '@brightspace-ui/core/components/demo/code-view.js';
+import 'playground-elements/playground-ide';
+import 'playground-elements/playground-code-editor';
```
margaree, line 3: Is this import needed in this file?

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+			playground-preview {
+				height: 100%;
+			}
+			.preview-container {
+				height: 100%;
+				position: relative;
+				border-radius: inherit;
+			}
+
+			.slider {
```
margaree, line 73: Ideally all these css attributes should be in alphabetical order. I think we would need to get the style linter running to catch these (I can add separately)

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+				background-color: white;
+				border-left: 1px solid #eff3fa;
+				z-index: 9;
+				cursor: col-resize;
+				display: flex;
+				align-items: center;
+				justify-content: center;
+			}
+			.slider:hover {
+				background-color: var(--d2l-color-gypsum);
```
margaree, line 88: An `import '@brightspace-ui/core/components/colors/colors.js';` should probably be added to the top (it's likely being pulled in from another file but just in case it got removed from wherever that was)

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+				-webkit-overflow-scrolling: touch;
+				overflow-x: hidden;
+				position: relative;
+			}
+		`;
+	}
+	constructor() {
+		super();
+		this.resizable = true;
+		this.attached = true;
```
margaree, line 107: nit: alphabetical order

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+		this.size = 'small';
+	}
+	firstUpdated() {
+		this._previewContainer = this.shadowRoot.querySelector('.preview-container');
```
margaree, line 111: Is this used?

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+			*/
+			imports: { type: String },
+			/**
+			* Is the preview resizable
+			*/
+			resizable : { type: Boolean },
+			/**
+			*  Is a code editor attached to the preview
+			*/
+			attached : { type: Boolean },
```
margaree, line 30: nit: these properties should be alphabetized

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,211 @@
...
+
+		const onPointerup = () => {
+			bar.releasePointerCapture(pointerId);
+			bar.removeEventListener('pointermove', onPointermove);
+			bar.removeEventListener('pointerup', onPointerup);
+		};
+		bar.addEventListener('pointerup', onPointerup);
+	}
+}
+customElements.define('d2l-component-catalog-demo-resizable-preview', DemoResizablePreview);
```
margaree, line 211: nit: for consistency `DemoResizablePreview` should be `ComponentCatalogDemoResizablePreview`

pages/assets/demo-snippet.js
```diff
@@ -0,0 +1,124 @@
...
+				display: none;
+			}
+
+			playground-code-editor {
+				min-width: ${MINIMUM_WIDTH}px;
+				width: 100%;
+				display: inline-block;
+				--playground-code-font-family: 'Lato', 'Lucida Sans Unicode', 'Lucida Grande', sans-serif;
+				--playground-code-background: var(--d2l-color-ferrite);
+				/* todo: do some of these colors have names?? */
```
margaree, line 59: Looks like some do - can find those here https://github.com/BrightspaceUI/core/blob/master/components/colors/colors.js

.eleventy.js
margaree: I think a prop to only show the code section seems fine

rollup.config.js
```diff
@@ -5,8 +5,8 @@ import outputManifest from 'rollup-plugin-output-manifest';
 
 const componentFiles = [
 	'./pages/assets/base-imports.js',
-	'./pages/assets/code-view-wrapper.js',
-	'./pages/assets/demo-snippet-wrapper.js',
+	'./pages/assets/demo-resizable-preview.js',
```
margaree, line 8: This is also probably not needed if it's only used in `demo-snippet`

margaree (approved):
I tested this out with the `component-catalog-doc-changes` branch on core and it's looking really good! Something I noticed is that is that in the Button doc, with the top demo which has 4 buttons in it, if I scale the space really small the buttons don't wrap.

This looks like a good start to me. Other than the latest changes mentioned, I'm good for other pieces/tweaks to go in in separate PRs if you'd like.

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,194 @@
...
+				background-color: var(--d2l-color-gypsum);
+			}
+			#preview::part(preview-toolbar) {
+				display: none;
+			}
+		`;
+	}
+	constructor() {
+		super();
+		this.size = 'small';
```
dlockhart, line 92: Should be setting defaults for `attached` and `resizable` to `false` here.

pages/assets/demo-resizable-preview.js
dlockhart: The eventual style linter is also going to complain about the CSS class names -- it'll want `d2l-*` prefixes to avoid collisions with other CSS on the page (although now shadow DOM makes this a non-issues everywhere except legacy-Edge).

pages/assets/demo-resizable-preview.js
```diff
@@ -0,0 +1,194 @@
...
+		};
+		const previewStyles = {
+			width: `calc(100% - ${this.resizable ? SLIDER_WIDTH : 0}px)`,
+			borderRadius: `10px ${resizableBorderRadius} ${resizableBorderRadius} ${attachedBorderRadius}`
+		};
+		const previewContainerStyles = {
+			width: this._previewWidth ? `${this._previewWidth}px` : '100%',
+			maxWidth: '100%', // prevent preview from expanding past code block when page resizes
+			minWidth: `${MINIMUM_WIDTH}px`
+		};
```
dlockhart, line 112: You may be able to replace a lot of this by using CSS attribute selectors in your CSS. For example:

```css
.slider {
  border-radius: 10px;
}
:host([attached]) .slider {
  border-radius: 0;
}
```

Note that the property definitions will need `reflect: true` for that to work.

dlockhart (approved):
A few more comments, but I'm also good with this going in and I'm sure there'll be more tweaks over time.

PR #144, opened 2021-06-02
Status Table Page

margaree:
This PR adds the following:
- Status table page with tables for Official, Labs, and Request components (based on labels on issues - modifies the issue templates as needed)
- Each component has a page. If `devMarkdown` is included (meaning the component has complete documentation) then that is used as the page, else generate a page from the GitHub issue body
- We aren't going to use design status at the moment - might be added later.


pages/components/status.md
```diff
@@ -0,0 +1,35 @@
...
+eleventyNavigation:
+  title: Status
+  key: status
+  parent: components
+  order: 1
+---
+
+# Component Status
+
+This page lists all our components, including [Labs]() components and even [requested]() components that haven't been built yet. We include the overall status of the component, but you can get the full details by clicking on the component name or its status.
```
margaree, line 12: Links currently don't go anywhere yet.

tools/generate-pages.mjs
```diff
@@ -91,6 +85,58 @@ function _getIssues(issues, expectedLabelName) {
...
 }
 
+function _generatePage(frontMatter, content) {
+	Object.keys(frontMatter).forEach(key => frontMatter[key] === undefined && delete frontMatter[key]); // remove any keys with an undefined value else matter.stringify fails
+	const newFile = path.join(DIR_IMPORTED_COMPONENTS, `${frontMatter.fileName}.md`);
+	const frontMatterString = matter.stringify('', frontMatter);
+	fs.writeFileSync(newFile, `${frontMatterString}\n${content}`);
+}
+
+function _getDevStatus(labels, state, issueState) {
```
margaree, line 95: For getting the dev status I added "Deprecated" and "Backlog" labels since something can have Complete development but be deprecated or something can have a not started dev state but be on the backlog or not planned.

I added the fallback assumption of complete or not started in case there are typos. We can also add labels for "Development In Progress" etc but my worry would be that if we bring design statuses back in, there could be an excessive amount of labels.

tools/generate-pages.mjs
```diff
@@ -1,3 +1,4 @@
+import { DEV_STATES, ISSUE_LABELS } from './states.mjs';
```
margaree, line 1: This file was getting overly complicated, especially with the generating of pages from issues that don't have documentation, so I tried to do some rearranging to make it a bit more straightforward.

tools/states.mjs
```diff
@@ -0,0 +1,24 @@
+export const CC_STATES = {
```
margaree, line 1: I split these all out in case they have changes in terminology - `CC_STATES` are the states that show up in the status table (and could end up as a combination of dev and design states, though currently they are just determined by dev state), `DEV_STATES` are determined through the state in the issue comment and the labels on it.

GeorgePlukov (changes requested):

.eleventy.js
```diff
@@ -84,6 +89,20 @@ module.exports = function(eleventyConfig) {
 
 	const defaultTextRule = markdownIt.renderer.rules.text;
 	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
+		const content = tokens[idx].content;
+		if (env.tags && Object.keys(env.tags[0]).includes(content)) {
+			const tag = env.tags[0][content];
+			return `
+				${defaultTextRule(tokens, idx, options, env, slf)}
+				<div class="d2l-component-catalog-tag d2l-body-standard">
+					<<div class="d2l-component-catalog-tag-inner">${tag}</div>>
```
dlockhart, line 98: Looks like some extra `<` and `>` snuck in here? If they were intentional to show the tag name, probably better to use `&lt;` and `&gt;`.

pages/assets/status-table.js
```diff
@@ -0,0 +1,122 @@
...
+	} else {
+		return CC_STATES.NOT_PLANNED;
+	}
+}
+
+class ComponentCatalogStatusTable extends LitElement {
+
+	static get properties() {
+		return {
+			type: { type: String },
```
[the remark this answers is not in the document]
dlockhart: I'm fine not documenting components like this that are purely internal to the catalogue.

pages/assets/status-table.js
```diff
@@ -0,0 +1,122 @@
...
+		this._components = [];
+	}
+
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		if (!this.type) return;
+
+		this._components = components[this.type];
+		this._components.sort((a, b) => {
```
dlockhart, line 77: Would it make sense for the generated issue data to already be sorted alphabetically? Would save runtime client-side sorting over and over again.

pages/assets/status-table.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	render() {
+		if (!this._components || this._components.length === 0) return html`<div>There are currently no components of this type.</div>`;
+
+		const rows = this._components.map(component => {
+			const linkHref = component.fileName ? `../${component.fileName}.html` : component.issueUrl;
+			const name = this.type === 'official'
+				? html`<d2l-link href="${linkHref}">${component.name}</d2l-link>`
+				: html `<d2l-link class="d2l-component-catalog-status-table-name" href="${linkHref}">${component.name}<img src="/img/${this.type}-icon.svg" alt="${this.type} icon"></d2l-link>`;
```
dlockhart, line 91: `alt` on this image can be likely be null... the fact that it's labs/etc. should already have been captured by the heading above this table and likely doesn't need to be repeated with each one?

pages/assets/status-table.js
```diff
@@ -0,0 +1,122 @@
...
+					<td>${name}</td>
+					<td>${status}</td>
+					<td>${owner}</td>
+				</tr>
+			`;
+		});
+
+		return html`
+			<d2l-table-wrapper>
+				<table class="d2l-table">
```
dlockhart, line 108: This ends up being super clean, love it!

pages/components/status.md
```diff
@@ -0,0 +1,35 @@
...
+
+To learn more on how we organize components see [Component Tiers]().
+
+## Official Components
+The "official" tier is reserved for components that have complete buy-in from design and are common to many pages or workflows in Brightspace.
+
+Official components adhere to our architectural guidelines, are localized and highly performant. They have a full suite of automated tests (include accessibility and perceptual diff), a demo page, and appropriate documentation.
```
dlockhart, line 19: I know you just lifted this from the Guide, but we've coalesced around "visual diff" so I think we should use that terminology here. I've updated the Guide to do the same.

```suggestion
Official components adhere to our architectural guidelines, are localized and highly performant. They have a full suite of automated tests (including accessibility and visual diff), a demo page, and appropriate documentation.
```

pages/components/status.md
```diff
@@ -0,0 +1,35 @@
...
+
+{% statusTable "official" %}
+
+## ![labs icon](/img/labs-icon.svg) Labs Components
+Components that fall into the "labs" tier have some buy-in from design, at least in an experimental sense. They may not have a full pattern documented on design.d2l yet, but some design thought has gone into the component and where and how it should be used. The component has the potential to be used in multiple places.
```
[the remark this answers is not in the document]
dlockhart: Margaree's mostly borrowing from [this page in the Guide](https://github.com/BrightspaceUI/guide/wiki/Component-Tiers). I agree though that "buy-in" is a little weird. "Support"?

Agree that we likely don't want to mention design.d2l here since the Catalogue will in all likelihood be replacing it:

```suggestion
Components that fall into the "labs" tier have some support from design, at least in an experimental sense. Some design thought has gone into the component and where and how it should be used. The component has the potential to be used in multiple places.
```

Another option would be to exclude this stuff entirely from this particular page and just show the tables, and then we move all the info about tiers -- why we have them, what they are, etc. -- into the secondary doc that you linked to above on Component Tiers.

pages/components/status.md
```diff
@@ -0,0 +1,35 @@
...
+
+Labs components may not fully adhere to archutectural guidelines, and may be missing some tests or documentation.
```
dlockhart, line 26:
```suggestion
Labs components may not fully adhere to architectural guidelines, and may be missing some tests or documentation.
```

pages/components/status.md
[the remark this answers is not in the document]
margaree: So the wording in the status page is pretty much just directly copied from Jeff's design for the page. I like the suggestions though. I'll make them and run them by him to confirm.

.eleventy.js
margaree: They are intentional to show the tag. `&lt;` and `&gt;` do seem better - will change.

pages/components/status.md
margaree: 🤦‍♀️

GeorgePlukov (approved):

PR #145, opened 2021-06-04
linting: add stylelint and lit-analyzer

margaree:


GeorgePlukov (approved):
I'm surprised there wasn't more linting issues 👍

dlockhart:
Hmm I'm surprised too... I would have expected the non-d2l prefixed CSS classes to fail at least. You sure it's running on the components?

margaree:
I think all the CSS classes included so far that this would have run on are prefixed with `d2l`

dlockhart:
Ah yes, because @GeorgePlukov's PR isn't merged yet, makes sense!

PR #148, opened 2021-06-09
Deploy: try assuming role first

margaree:


.github/workflows/deploy.yml
```diff
@@ -23,10 +23,17 @@ jobs:
         run: npm i && npm i aws-sdk --no-save
       - name: Build
         run: npm run build:prod
+      - name: Assume role
+        uses: Brightspace/third-party-actions@aws-actions/configure-aws-credentials
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
+          role-to-assume: arn:aws:iam::037018655140:role/daylight.d2l.dev-githubactions-access
```
svanherk, line 32: This _might_ need to be in quotes, I'm not sure

svanherk (approved):

.github/workflows/deploy.yml
margaree: So I did a search https://search.d2l.dev/search?project=Brightspace&project=BrightspaceUI&full=%22role-to-assume%22&defs=&refs=&path=&hist=&type=&xrd=&nn=2&si=full&si=full and it looks like other places don't so hopefully it works

PR #149, opened 2021-06-09
Deploy: branches and commits

margaree:
Branches:
- Syncs to `.../branches/${branchName}` anytime there is a push to a branch (would include master)
- Note: From some googling it looks like this `${GITHUB_REF##*/}` will get just the branch name out of the ref [update: this piece did work since it ran on this branch]

Commits:
- Syncs to `.../branches/${commit sha}` anytime there is a push to master

Question: am I using the quotes and curly brackets correctly? I tried to match examples I saw so I think this would work, but I could also seeing it being an overuse of quotes.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,5 +1,8 @@
 name: Deploy
-on: push
+on:
+  push:
+    branches:
+      - master
 jobs:
   deploy:
     name: Deploy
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -32,8 +29,18 @@ jobs:
           role-to-assume: arn:aws:iam::037018655140:role/daylight.d2l.dev-githubactions-access
           role-duration-seconds: 3600
           aws-region: ca-central-1
-      - name: Deploy
-        run: aws s3 sync --delete ./_site/ s3://daylight.d2l.dev/branches/master
+      - name: Deploy - branch
+        run: |
+            S3_TARGET="s3://daylight.d2l.dev/branches/${GITHUB_REF##*/}"
```
svanherk, line 34: Are there characters we can have in a branch name that wouldn't work in a bucket path?

margaree: That's a good point - their ability contain slashes also makes it a bit weird. Since we want to use these for PRs it seems like it might be safer to use the PR number and have this run on pull request.

svanherk: Yeah this workflow isn't very complicated so I don't think a bit of duplicate logic would be bad if you wanted to separate out a `push to master` workflow and a `pr` workflow.  Then you could use something like https://github.com/actions/github-script#comment-on-an-issue to comment the preview link on the PR without having to check we're in a PR first, etc

svanherk: Or I guess you could just add to the existing `ci` one, and change this back to push on master

PR #150, opened 2021-06-09
Interactive table for demo components

GeorgePlukov:


pages/assets/demo-snippet.js
```diff
@@ -109,20 +134,47 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 	}
 
 	get imports() {
-		return getImports(this.demoSnippet);
+		return parseImports(this.demoSnippet);
+	}
+	get size() {
+		const tag = 'size:';
```
GeorgePlukov, line 140: Does this seem like a reasonable way to handle parsing additional parameters from the demos? I think we'll need to add at least one or two more related to positioning of the component within the iFrame to allow the consumer to override the `justify-content`  and `align-items` flex properties.

Right now they are set to 

```css
justify-content: space-evenly;
align-items: center;
```

pages/assets/demo-attribute-table.js
```diff
@@ -0,0 +1,177 @@
...
+			'property-change',
+			{ bubbles: true, composed: false, detail: details }
+		));
+	}
+
+	_getDemoValueOptions(type, attributeName, defaultVal) {
+		const value = defaultVal ? defaultVal.replace(/"/g, '') : undefined;
+		switch (type) {
+			case 'array':
+			case 'object':
```
GeorgePlukov, line 105: I'm not sure if we want to be able to edit object properties. The ones I could find are listed here: 

`Tooltip/dropdown content boundaries`: This could be useful but I couldn't find a great way to document what the accepted object properties were here ( in this case \"top\", \"bottom\", \"left\", and \"right\") in order to pass them through to the `custom-elements` file

`Input date` - `child errors`: @margaree maybe you could comment on whether this one should be editable?

`d2l-form-error-summary`: Would we have a specific demo for this? I haven't looked too much into what we are doing with forms

margaree: For tooltip I did keep the section that talks about the boundaries so in the future we could add a link to that or just put "See below" since that piece is complicated.

We wont want `child errors` to show up for `input-date` (I think we potentially would want to not be showing anything `FormElementMixin` related there). I would say ignore anything forms related for now - that piece will be a bit different than the others and will need additional thought.

pages/assets/demo-resizable-preview.js
```diff
@@ -143,6 +150,9 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			</body>`;
 	}
 	firstUpdated() {
+		this._previewContainer = this.shadowRoot.querySelector('.preview-container');
```
margaree, line 153: `_previewContainer` doesn't seem to be used

pages/assets/demo-resizable-preview.js
```diff
@@ -185,7 +196,19 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			</div>
 		`;
 	}
+	update(changedProperties) {
```
margaree, line 199: Should this be `updated`?

pages/assets/demo-snippet.js
margaree: Do you picture this being used like:
`<!-- docs: live demo size: small -->`
or like:
`<!-- docs: live demo --> size: small ... `
or something completely different?

pages/assets/demo-snippet.js
```diff
@@ -1,27 +1,30 @@
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/demo/demo-snippet.js';
 import './demo-resizable-preview.js';
+import './demo-attribute-table.js';
```
margaree, line 4: nit: alphabetical order

pages/assets/demo-snippet.js
GeorgePlukov: I was imagining it being used as your first example along with `name:d2l-button` for whatever attributes we would be pulling. Technically the way it works now it doesn't matter where in the demo snippet the tags exist.

pages/assets/demo-attribute-table.js
```diff
@@ -0,0 +1,177 @@
...
+	}
+	static get styles() {
+		return [heading4Styles, selectStyles, tableStyles, css`
+			:host {
+				display: block;
+			}
+			:host([hidden]) {
+				display: none;
+			}
+			td.d2l-table-cell-first,
```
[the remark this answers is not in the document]
GeorgePlukov: These classes are added by d2l-table for the first cell in a row. It prevents long/hyphenated attribute names from wrapping to a second line

pages/assets/demo-attribute-table.js
```diff
@@ -0,0 +1,177 @@
...
+	}
+	_getEventDetails(e, type) {
+		const value = type === 'Boolean' ? e.target.hasAttribute('on') : e.target.value;
+		return {
+			name: e.target.getAttribute('data-name'),
+			type: type,
+			value
+		};
+	}
+	_onNumberChange(e) {
```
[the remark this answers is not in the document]
GeorgePlukov: Our linting rules prevent arrow functions and binding within templates so I don't think this is possible/ideal here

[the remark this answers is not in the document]
margaree: ah interesting, okay nevermind!

pages/assets/demo-resizable-preview.js
GeorgePlukov: I left this as updated so that the demo-previews content gets set before the render cycle happens. It shouldn't really matter I don't think

pages/assets/demo-snippet.js
GeorgePlukov: Sorry I missed this comment last week somehow. I definitely think this needs to be set per demo instance so I don't think we can have it live on the issue level. I don't mind the object formatting but it does add a lot of extra characters to be typing out every time we are adding a demo.  Do you think formatting as an object is worth adding in this PR? Maybe we take a look at it if we have trouble with keeping the demo comments organized?

pages/assets/demo-attribute-table.js
```diff
@@ -0,0 +1,186 @@
...
+						label="${attributeName}"
+						label-hidden
+						value="${ifDefined(value)}">
+					</d2l-input-text>`;
+			default: {
+				// the case of an array of strings
+				let options = type.replace(/'/g, '').split(' | ');
+
+				// Add empty default values to the dropdown lists
+				if (!options.includes(strippedDefaultVal)) {
```
GeorgePlukov, line 141: I added this so that the select drop downs also get populated with an empty item in the case where the default isn't included in the options list

pages/assets/demo-attribute-table.js
```diff
@@ -0,0 +1,177 @@
...
+			if (info.type && info.type.includes('|')) {
+				demoType = info.type.replace(/\|/g, ' | ');
+				demoValue = this._getDemoValueOptions(demoType, info.name, infoDefault);
+			} else {
+				demoValue = validTypes.includes(demoType) ? this._getDemoValueOptions(demoType, info.name, infoDefault) : null;
+			}
+			const demoValueRow = this.interactive ? html`<td>${demoValue}</td>` : null;
+			return html`
+				<tr>
+					<td>${info.name}</td>
```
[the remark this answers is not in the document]
GeorgePlukov: The table header is defined a little further down in this file, these are actually rows being generated for the table contents

pages/assets/demo-snippet.js
margaree: Would this still work if the comment is over multiple lines, e.g.:
```
<!-- docs: live demo
name: d2l-button
-->
...
```
The one line just could really easily end up looking quite cluttered and hard to read.

pages/assets/demo-attribute-table.js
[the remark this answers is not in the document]
dlockhart: Yep totally -- `th` elements are valid for rows as well! They provide screen readers with row scope in addition to column scope when they're navigating the table cells.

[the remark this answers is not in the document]
GeorgePlukov: TIL! I added `scope="row"` as well, I think that is the correct usage. Thanks for pointing this out!

pages/assets/demo-snippet.js
GeorgePlukov: I added support for both formats! Being able to separate lines should be handy

pages/assets/demo-resizable-preview.js
```diff
@@ -112,13 +114,18 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 	}
 	get indexHTML() {
 		return `
-			<script type="module" src="index.js"></script>
 			<script>
 				window.addEventListener('load', function () {
 					var demoEl = document.getElementById('demo-element');
 					demoEl.classList.remove('hide');
 				});
+				// Suppress errors only in production? This will hide any errors with attributes and the module not resolved errors
```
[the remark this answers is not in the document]
margaree: Would you mind creating an issue so that we remember to deal with this/uncomment this code?

margaree (approved):

PR #152, opened 2021-06-11
Deploy: Add comment to PR with preview url

margaree:


.github/workflows/ci.yml
```diff
@@ -36,3 +36,14 @@ jobs:
         env:
           # cred variables set in the "Assume role" step
           AWS_DEFAULT_REGION: ca-central-1
+      - uses: Brightspace/third-party-actions@actions/github-script
+        if: github.event.action == 'opened'
```
margaree, line 40: I added this so that the comment is only added when the PR is opened and not each time changes are pushed to the PR.

margaree:
Okay so this PR has been changed so that the Preview URL has been added as a neutral-resulting check, and the comment will no longer be added. The URL that appears there can't be made a link that I've found (though if you click into details it's a bit nicer) and also I can't change "Details" to point to it (despite the API having a `details_url` - doesn't seem to work, but copy/pasting the link isn't awful. Let me know any thoughts on this!

svanherk:
So I like either (the check is nice because it's one less email, and the comment is nice because it's a quick click versus copy/paste)!  I kinda miss the row of checkmarks, but it _seems_ like the neutral check is maybe stopping the checks from from auto-collapsing (which would be a good thing here)

dlockhart:
Agree with Stacey here -- both are good for different reasons. Also good with either.

With the status check, how much extra functionality there do they expose? Like can its state be "unknown"/"error"/"success"? If so -- and we wouldn't do this now -- but the link could launch the site in a little wrapper with a bar at the top where you could click "Approve" that would move the status from unknown -> success. Kind of like reviewing visual diffs.

margaree:
When I was looking into the checks there is the concept of `actions` where you can have a button to do something so I think in theory we could do something like that which would be useful. There is an `action_required` state which I think would indicate be useful there (though its icon is a red X which I don't necessarily agree with).

dlockhart:
Cool yeah... probably similar to the old visual diff stuff. I think if you write an actual app integration you can do more things but that's a lot.

Good with either of these approaches in the meantime, whichever you prefer!

PR #156, opened 2021-06-14
Unit Tests

margaree:


.eleventy.js
```diff
@@ -8,7 +8,8 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addPassthroughCopy('pages/favicon.ico');
 	eleventyConfig.addPassthroughCopy('pages/img');
 	eleventyConfig.addNunjucksShortcode('enhancements', (repo) => {
-		return `Looking for an enhancement not listed here? <d2l-link href="${repo}/issues">Create a GitHub issue!</d2l-link>`;
+		const href = repo ? `${repo}/issues` : 'https://github.com/BrightspaceUI/documentation/issues';
```
margaree, line 11: I switched this so that in case there wasn't a repo/repo was undefined, there would still be a URL. Another option is to not include this line if no repo.

.eleventy.js
```diff
@@ -93,6 +94,7 @@ module.exports = function(eleventyConfig) {
 	const defaultTextRule = markdownIt.renderer.rules.text;
 	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
+		// Note: if tags are formatted like eleventyNavigation then we get error "entry.data.tags is not iterable" when parsing
```
margaree, line 97: I added this note in case future me thinks "why don't we do it this way" for the third time.

tools/generate-pages-utils.mjs
```diff
@@ -0,0 +1,66 @@
+import { DEV_STATES, ISSUE_LABELS } from './states.mjs';
```
margaree, line 1: These are just moved from `generated-pages.mjs` so they can be tested (if they're kept in there then the whole file also executes)

tools/generate-pages-utils.mjs
```diff
@@ -0,0 +1,66 @@
...
+		comment = '';
+	const splitStart = issueBody.split(/<!--\r?\n/);
+	if (splitStart.length === 2) {
+		const splitEnd = splitStart[1].split('-->');
+		comment = splitEnd[0];
+		issueBody = splitEnd[1];
+	}
+
+	const title = issue.title.replace(/\s+/g, '-').toLowerCase();
+	let info = {
```
margaree, line 40: I did remove `info.name` and `frontMatter.repo = info.repo` since (a) we used `info.name` once and can just use `frontMatter.title` for the case where that was needed, and (b) we can just put the repo in the frontMatter in the GitHub issues since it doesn't need to be in info.

tools/generate-pages.mjs
```diff
@@ -36,7 +39,8 @@ function _copyCustomElements(repos) {
 	repos.forEach((repo) => {
 		const customElementsFilePath = path.join(__dirname, `../node_modules/${repo}/custom-elements.json`);
 		if (!fs.existsSync(customElementsFilePath)) {
-			console.warn(`WARNING: custom-elements.json does not exist for ${repo}`);
+			const message = `custom-elements.json does not exist for ${repo}`;
+			_handleError(message);
```
margaree, line 43: I changed this and the lack of readme file to throw errors when prod so that those cases would get caught in CI.  Other cases in the files that caused warnings did not seem worth full failures to me but these ones seemed important (mostly the missing README file)

tools/generate-pages.mjs
```diff
@@ -139,6 +130,13 @@ function _getMissingDependenciesList(dependencies) {
 	});
 }
 
+function _handleError(message) {
+	if (PROD) {
+		console.error(`ERROR: ${message}`);
+		process.exit(5);
```
margaree, line 136: Error code 5 corresponds to Fatal Error (https://nodejs.org/api/process.html#process_exit_codes) which seemed right. I'm not super familiar with error code selection though so definitely open to suggestion here.

tools/generate-pages.mjs
```diff
@@ -149,64 +147,13 @@ function _installDependencies(dependencies, callback) {
 	});
 }
 
-function _parseBody(issue) {
```
margaree, line 152: Moved to the util file

dlockhart (approved):

.eleventy.js
GeorgePlukov: Would someone be clicking this link if they had an idea to improve the documentation for a certain component/web page? If so I think it makes sense to link them here

tools/generate-pages.mjs
GeorgePlukov: I think the if we want this to fail here we should just treat it as an uncaught error with exit code 1. I think exit code 5 is typically reserved specifically for V8 errors.

GeorgePlukov (approved):
Left a couple comments but looks good!

.eleventy.js
margaree: That could happen, yup.

PR #158, opened 2021-06-18
Use importMetaAssets to resolve service workers for playground-elements

GeorgePlukov:
Added some required compilation steps for bundling (https://github.com/PolymerLabs/playground-elements#bundling).


margaree (approved):
Tried this out and all looked good to me👍

PR #159, opened 2021-06-21
Visual diff tests: Setup & Status Table tests

margaree:


.eleventy.js
```diff
@@ -4,9 +4,9 @@ const eleventyNavigationPlugin = require('@11ty/eleventy-navigation');
 const { escapeHtml } = require('markdown-it/lib/common/utils');
 
 module.exports = function(eleventyConfig) {
+	eleventyConfig.addPassthroughCopy('img');
```
margaree, line 7: I moved img so that the images could be picked up by the visual-diff tests

test/status-table.visual-diff.html
```diff
@@ -0,0 +1,103 @@
...
+					{
+						'name': 'Not Started with Team',
+						'development': 'Not Started',
+						'owner': 'Great Team'
+					}
+				]
+			};
+
+			window.addEventListener('DOMContentLoaded', () => {
+				const tables = document.querySelectorAll('d2l-component-catalog-status-table');
```
margaree, line 69: This section is for setting the components data. I couldn't figure out how to mock the file returned but this worked.

GeorgePlukov (approved):
LGTM. Do you think we'll try and add some sort of visual diff testing for the live site (not mocked data)? I know there will be edits happening in the README's but might be a nice way to check what is changing on the site before deploys.

margaree:
Yup agreed something like that would be nice. I'm going through the components first and then hoping to look into something bigger picture.

PR #161, opened 2021-06-21
Fix cache

margaree:
I also added caching to ci and visual-diff


.github/workflows/ci.yml
```diff
@@ -10,6 +10,11 @@ jobs:
       - uses: Brightspace/third-party-actions@actions/setup-node
         with:
           node-version: '14'
+      - uses: Brightspace/third-party-actions@actions/cache
+        id: cache
+        with:
+          path: '**/node_modules'
+          key: npm-${{ hashFiles('**/package.json') }}
```
dlockhart, line 17: I'm not sure if just using `package.json` as the hash is going to work. In theory because it has lots of `^1` type things in it, `npm install` could install new things from the previous run but there'd still be a cache hit so those new versions of things wouldn't get used. This is why we ended up needing to add a lock file to core.

dlockhart: Looks like you fixed that already! :)

margaree: Yeah that's definitely a potential issue. I've switched it to cache on package-lock and added package-lock.

.github/workflows/deploy.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           key: npm-${{ hashFiles('**/package-lock.json') }}
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
-        run: npm i && npm i aws-sdk --no-save
+        run: npm ci && npm i aws-sdk --no-save
```
dlockhart, line 23: This may be buggy. The step is going to be skipped if there's a cache hit, but `aws-sdk` won't be in the cached `node_modules` if the thing that populated it was the visual-diff or ci workflow. I'd recommend adding a separate step below "Install dependencies" to install the extra ones.

dlockhart (approved):

PR #165, opened 2021-06-22
Add edit URL to component pages

GeorgePlukov:
Also fixed a typo in the demo-snippet in this PR.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -14,6 +14,13 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addNunjucksShortcode('issue', (issueUrl) => {
 		return `Looking for more details on the component or want to add your input? <d2l-link href="${issueUrl}">Check out the GitHub Issue</d2l-link>`;
 	});
+	eleventyConfig.addNunjucksShortcode('editPage', (repo, componentPath) => {
+		// ensure the repo or component path have a '/' before concatenation
+		if (repo.slice(-1) !== '/' && componentPath.slice(0) !== '/') {
+			repo += '/';
+		}
+		return `Suggest an <d2l-link href="${`${repo}edit/master/${componentPath}`}">edit</d2l-link> for this page`;
+	});
 	eleventyConfig.addShortcode('statusTable', (tier) => {
 		return `<d2l-component-catalog-status-table tier="${tier}"></d2l-component-catalog-status-table>`;
 	});
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -22,7 +22,7 @@ module.exports = function(eleventyConfig) {
 		if (repo.slice(-1) !== '/' && componentPath.slice(0) !== '/') {
 			repo += '/';
 		}
-		return `<div class="d2l-edit-component">Suggest an <d2l-link href="${`${repo}edit/master/${componentPath}`}">edit</d2l-link> for this page</div>`;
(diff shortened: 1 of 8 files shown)


pages/_includes/layouts/component.njk
```diff
@@ -7,3 +7,5 @@ permalink: components/{{fileName}}.html
 
 <h2 class="d2l-heading-2">Future Enhancements</h2>
 {% enhancements repo %}
+
+<h5 class="d2l-heading-5">{% editPage repo, devMarkdown %}</h5>
```
margaree, line 11: I don't think this needs to be any sort of heading. Generally with headings order is important, i.e., this one would be an h3 if under Future Enhancements and then if we wanted it to look different we'd add a class and style it differently. This can likely just be a div block though.

margaree (commented):
You'll also want to adjust the places in the `generate-pages` script to use `frontMatter.devMarkdown` (e.g., https://github.com/BrightspaceUI/documentation/blob/master/tools/generate-pages.mjs#L89). Maybe adjust the button issue to only have devMarkdown and repo only in the frontMatter to confirm it works as expected.

pages/_includes/layouts/component.njk
GeorgePlukov: I was thinking this could be its own sort of mini section that was just a header, but mostly just highjacking the styling. I switched it to using a div, Jeff might have some better designs for us at some point as I didn't see this in the XD's I have.

.eleventy.js
```diff
@@ -14,6 +14,16 @@ module.exports = function(eleventyConfig) {
...
 	});
+	eleventyConfig.addNunjucksShortcode('editPage', (repo, componentPath) => {
+		if (!repo || !componentPath) {
+			return '';
+		}
+		// ensure the repo or component path have a '/' before concatenation
+		if (repo.slice(-1) !== '/' && componentPath.slice(0) !== '/') {
+			repo += '/';
+		}
+		return `<div class="d2l-edit-component-page">Suggest an <d2l-link href="${`${repo}edit/master/${componentPath}`}">edit</d2l-link> for this page</div>`;
```
margaree, line 25: Something we might need to be aware of is that there could be repos that use `main` instead of `master`, though we can leave that for now.

margaree (approved):

.eleventy.js
GeorgePlukov: Good point, if/when we run into this I think  maybe we just add an override url for those specific components

.eleventy.js
```diff
@@ -14,6 +14,16 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addNunjucksShortcode('issue', (issueUrl) => {
 		return `Looking for more details on the component or want to add your input? <d2l-link href="${issueUrl}">Check out the GitHub Issue</d2l-link>`;
 	});
+	eleventyConfig.addNunjucksShortcode('editPage', (repo, componentPath) => {
+		if (!repo || !componentPath) {
+			return '';
+		}
+		// ensure the repo or component path have a '/' before concatenation
+		if (repo.slice(-1) !== '/' && componentPath.slice(0) !== '/') {
```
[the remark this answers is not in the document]
GeorgePlukov: Moved the code to parse the repo URL to the .eleventy file, ensures we don't have to duplicate the code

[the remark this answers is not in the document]
margaree: Yay for more unit tests too!🎉

margaree (approved):

PR #166, opened 2021-06-23
Switch subtitle parsing to use square brackets

GeorgePlukov:
This would allow us to remove the tags included in the github issue and parse the subtitle based on square brackets: `[subtitle-text]`. I was thinking maybe it is just best to use the formatting `<subtitle-text>` since that is essentially what is going to get rendered in the end. 

Let me know what you think of this approach, I read through a bit of documentation on markdownIt but not sure if its a best practice to edit the tokens.content. 

This also allows us to still include links as it doesn't match with `[text](link)`

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -94,14 +94,11 @@ module.exports = function(eleventyConfig) {
 	const defaultTextRule = markdownIt.renderer.rules.text;
 	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
-		// Note: if tags are formatted like eleventyNavigation then we get error "entry.data.tags is not iterable" when parsing
 		const subTitleMatch = content.match(/\[(.*?)\]/);
 		if (subTitleMatch) {
 			const subtitleMarkdown = subTitleMatch[0];
-			// remove the squire brackets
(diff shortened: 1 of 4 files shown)


.eleventy.js
```diff
@@ -95,8 +95,13 @@ module.exports = function(eleventyConfig) {
 	markdownIt.renderer.rules.text = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
 		// Note: if tags are formatted like eleventyNavigation then we get error "entry.data.tags is not iterable" when parsing
-		if (env.tags && Object.keys(env.tags[0]).includes(content)) {
-			const tag = env.tags[0][content];
+		const subTitleMatch = content.match(/\[(.*?)\]/);
+		if (subTitleMatch) {
+			const subtitleMarkdown = subTitleMatch[0];
+			// remove the squire brackets
```
margaree, line 101: nit: might look cleaner to have these comments on the same line. Also probably want `square`

margaree (approved):

margaree:
I think it's probably fine to just modify the `content`. Is there any concern if there are areas where someone just uses square brackets (or with the <>, just uses those - I think those would be more likely)? We could also maybe move this to be only in a heading.

margaree:
Also the tags section can now be removed from the `component-documented` template🎉

PR #167, opened 2021-06-23
Add styles for tables and code blocks

GeorgePlukov:
Changed the table styling to `light` and overrode some other styles to get the desired look. It seemed like it was going to be less css to change the `light` mode rather than the default styling


pages/assets/demo-attribute-table.js
```diff
@@ -25,13 +25,45 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
 	static get styles() {
 		return [heading4Styles, selectStyles, tableStyles, css`
 			:host {
+
```
margaree, line 28: nit: unnecessary line

margaree (commented):
Looks good! I think there should be a bit more padding on the left.
Also, this is only being applied to the demo attribute table. It would be nice if it could somehow be css shared between all the tables on the site (e.g., also status table, and any tables that would be included in a README like the accessibility properties)

pages/assets/demo-attribute-table.js
```diff
@@ -25,13 +25,45 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
...
+				font-weight: bold;
+			}
+			
+			th.d2l-table-cell-last {
+				border-radius: 0 10px 0 0;
+			}
+
+			th.d2l-table-cell-first {
+				border-radius: 10px 0 0 0;
+			}
```
dlockhart, line 52: All these overrides are definitely going to be fragile, and weren't really what we/I wanted consumers to be able to do with tables -- although there's not much I could do to prevent it. I'm mostly worried about future tweaks to table styling breaking this, and also designers/devs seeing it and copying it in their our UIs.

It might not actually be too difficult to just copy the light table styles from core and remove all the stuff that doesn't apply (sticky headers, row selection, etc.) and then add your tweaks.

.eleventy.js
```diff
@@ -91,10 +91,10 @@ module.exports = function(eleventyConfig) {
 	};
 
 	markdownIt.renderer.rules.table_open = () => {
-		return '<d2l-table-wrapper><table class="d2l-table">';
+		return '<table class="d2l-markdown-table">';
 	};
 	markdownIt.renderer.rules.table_close = () => {
-		return '</table></d2l-table-wrapper>';
+		return '</table>';
```
dlockhart, line 97: I believe this is going to replace _any_ table in people's markdown with these custom styles ones. Is that the intent? I don't think we use the table Markdown syntax very often anyway. 🤷

pages/assets/demo-attribute-table.js
```diff
@@ -105,22 +74,20 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
...
-							<th>Description</th>
-							<th>Default</th>
-							${demoValueHeading}
-						</tr>
-					</thead>
-					<tbody>
-						${rows}
-					</tbody>
-				</table>
-			</d2l-table-wrapper>
```
dlockhart, line 91: If you'd still like for the scroll buttons to appear when this table overflows, you can import and wrap it in `<d2l-scroll-wrapper>`.

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
+/* stylelint-disable */
+import { css } from 'lit-element';
+
+export const customTableStyles = css`
```
dlockhart, line 4: Since these styles are getting dropped into the root document, you may want to prefix them with a CSS class... otherwise any `<table>` that gets used anywhere in the site is going to be styled this way, and it kind of makes your `d2l-status-table`, `d2l-attribute-table` and `d2l-markdown-table` CSS classes pointless, no?

Since there's just one style here, I'd probably pick one CSS class and use it for all three and then later if variations on that style crop up we can adjust. Then these styles should become something like:

```css
table.d2l-whatever-table {
}
table.d2l-whatever-table tr > td {
}
```

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+		border: 1px solid #dfe6ef;
+		border-radius: 0.3rem;
+		border-spacing: 0;
+		width: 100%;
+	}
+
+	tr > td,
+	tr > th {
+		font-weight: inherit;
+		padding: 12px 6px 12px 6px;
```
dlockhart, line 15: Nit: shorthand is `padding: 12px 6px`

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+		text-align: left;
+		vertical-align: middle;
+	}
+
+	/* table > tbody > tr:nth-child(1) > th */
+	tbody:first-child > tr > th {
```
dlockhart, line 21: I don't quite get what this CSS rule is trying to target? The header row? Wouldn't something like this be clearer:

```css
table.d2l-whatever-table > thead > tr > th {
}
```

I guess unless the markdown tables don't actually generate a `thead`/`tbody`?

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+
+	/* override background color or headers that live within rows */
+	tbody > tr > th {
+		background-color: white;
+	}
+
+	/* padding for extra space on the left table edges */
+	tbody > tr > td:first-child,
+	tbody > tr > th,
+	thead > tr > th:first-child {
```
dlockhart, line 34: I think you can just leave the tag name off:

```suggestion
	tbody > tr > *:first-child {
```

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	td {
+		background-color: white;
+		border-bottom: 1px solid #dfe6ef;
+		line-height: 1rem;
+	}
+
+	/* round top left corner */
+	thead:first-child tr:first-child th:first-child {
```
dlockhart, line 61: There should be only one `thead`:

```suggestion
	table.d2l-whatever-table > thead > tr:first-child > th:first-child {
```

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+		border-radius: 0.3rem 0 0 0;
```
dlockhart, line 62: You can target the exact radius you want and avoid setting the others:
```suggestion
		border-top-left-radius: 0.3rem;
```

pages/assets/table-style.js
```diff
@@ -0,0 +1,77 @@
...
+	thead:first-child tr:first-child th:last-child {
+		border-radius: 0 0.3rem 0 0;
+	}
+	/* round bottom right corner */
+	tbody:last-child tr:last-child td:last-child {
+		border-radius: 0 0 0.3rem 0;
+	}
+	/* round bottom left corner */
+	tbody:last-child tr:last-child td:first-child,
+	tbody:last-child tr:last-child th:first-child {
```
dlockhart, line 74: Can be merged:

```suggestion
	table.d2l-table-whatever > tbody > tr:last-child > *:first-child {
```

pages/assets/table-style.js
GeorgePlukov: The intent behind the `d2-*-table` was more to make it obvious for us how the tables were being generated on the site (and leave it open for custom styling later). I thought it could be useful in differentiating between the markdown/component tables when inspecting the pages, so I am inclined to leave them for now. 

For now I think these styles should be applied to all tables but I'll add in a wrapper class since it's possible we want to style tables differently in some spots

dlockhart: Cool sounds good!

pages/assets/table-style.js
GeorgePlukov: Definitely should have been a `thead` here. The markdown tables do generate `thead`/`tbody` but originally I had my table formatted slightly incorrectly which was causing issues.  For the parser to properly add those tags the separator line between head/body in markdown has to have three dashes separating the bars (ex `|---|---|---|`) and it looks like some of our README's only have two (`|--|--|--|`).

Seems like most markdown previewers seem to recognize its a table with two dashes but not `markdown-it`, I'll make sure to right a note about that in our component transfer docs

pages/assets/status-table.js
```diff
@@ -100,8 +103,8 @@ class ComponentCatalogStatusTable extends LitElement {
 		});
 
 		return html`
-			<d2l-table-wrapper>
-				<table class="d2l-table">
+			<d2l-scroll-wrapper>
+				<table class="d2l-cc-custom-table d2l-status-table">
```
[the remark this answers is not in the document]
GeorgePlukov: I added those extra css classes just as a way to determine where the table is being generated from (markdown/status/component properties). I was thinking if we added other markdown tables here it would help distinguish between them for debugging

PR #169, opened 2021-06-24
Tests: visual-diff tests for demo-attribute-table & unit tests for demo-snippet

margaree:


pages/assets/demo-attribute-table.js
```diff
@@ -43,12 +44,16 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
 		`];
 	}
 
-	render() {
+	firstUpdated(changedProperties) {
+		super.firstUpdated(changedProperties);
+
+		this._componentInfo = components.find((component) =>  component.name === this.tagName);
```
margaree, line 50: Is there any issue with moving this here instead of having it in render? It makes mocking out the data easier and I think it would just need to happen once.

pages/assets/demo-snippet.js
```diff
@@ -6,31 +6,10 @@ import 'playground-elements/playground-code-editor';
 import 'prismjs/prism.js';
 import { css, html, LitElement } from 'lit-element';
 import { ifDefined } from 'lit-html/directives/if-defined.js';
+import { parseImports } from './utils.js';
 import { themeStyles } from './code-style.js';
 import { unsafeHTML } from 'lit-html/directives/unsafe-html';
 
-const defaultImports = [
```
margaree, line 12: I moved this out into a separate file for easier testing (there were some issues with exporting it from here and getting mocha to work)

pages/assets/demo-attribute-table.js
GeorgePlukov: I think this should be alright, I don't see a case where tag names would be changing

test/demo-attribute-table.visual-diff.html
```diff
@@ -0,0 +1,261 @@
...
+					{
+						"name": "items",
+						"description": "The items",
+						"type": "array"
+					},
+					{
+						"name": "type",
+						"description": "Type being displayed",
+						"type": "'default'|'critical'|'success'|'warning'",
+						"default": "\"\\\"default\\\"\""
```
GeorgePlukov, line 129: Is this formatting necessary? In the custom-elements I saw them formatted like `"default": "\"leftright\""`

GeorgePlukov (approved):

test/demo-attribute-table.visual-diff.html
margaree: Interesting - I'd grabbed that from `custom-elements-all` but looking at it again it's `"default": "\"default\""`. I'll update that.

PR #173, opened 2021-06-28
Add table for showing component slots

GeorgePlukov:


pages/assets/demo-attribute-table.js
```diff
@@ -76,6 +80,16 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
 				</tr>
 			`;
 		});
+
+		const slotRows = this._componentInfo.slots.map((slotInfo) => {
+			return html`
+			<tr>
+				<th scope="row">${slotInfo.name || 'Default'}</th>
```
GeorgePlukov, line 87: @margaree From the bit of googling I did seems like `default` or `unnamed` would fit here

pages/assets/demo-attribute-table.js
```diff
@@ -76,6 +80,16 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
 				</tr>
 			`;
 		});
+
+		const slotRows = this._componentInfo.slots.map((slotInfo) => {
+			return html`
+			<tr>
```
margaree, line 86: nit: this block should all be indented one further

pages/assets/demo-attribute-table.js
```diff
@@ -95,6 +109,20 @@ export class ComponentCatalogDemoAttributeTable extends LitElement {
 					</tbody>
 				</table>
 			</d2l-scroll-wrapper>
+			${!this.hideSlots ? html`<h3 class="d2l-heading-4">Slots</h3>
+			<d2l-scroll-wrapper>
```
margaree, line 113: This should also probably be indented further

margaree (commented):
Looks good! Questions:
1. Should the component be renamed since it's no longer just for an attribute table?
2. Should the slot name be code-styled (e.g., `default` rather than default)?
3. Can you add another visual-diff test for the case where the data doesn't contain slots but also doesn't have `hide-slots`?

GeorgePlukov:
> 1. Should the component be renamed since it's no longer just for an attribute table?
I was considering renaming but couldnt think of a great name. `ComponentCatalogDemoElementTables` or just `ComponentCatalogDemoTables` could work?
> 2. Should the slot name be code-styled (e.g., `default` rather than default)?
Good point, should be styled I think.
> 3. Can you add another visual-diff test for the case where the data doesn't contain slots but also doesn't have `hide-slots`?
👍

margaree (approved):

PR #181, opened 2021-06-29
Add default value option to demos, other tweaks

margaree:
This allows for demos to set default attribute values. Also some other tweaks I found when adding Button page.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -51,7 +51,7 @@ module.exports = function(eleventyConfig) {
 					let headingClass = token.attrGet('class') || '';
 					if (token.tag === 'h1') headingClass += ' d2l-heading-1';
 					else if (token.tag === 'h2') headingClass += ' d2l-heading-2';
-					else if (token.tag === 'h3') headingClass += ' d2l-heading-3';
+					else if (token.tag === 'h3') headingClass += ' d2l-heading-4';
 					token.attrSet('class', headingClass);
 					break;
 				}
--- a/docs/adding-component.md
+++ b/docs/adding-component.md
@@ -77,12 +77,13 @@ size:<'small'|'medium'|'large'|'xlarge'>
 
 **Interactive demo**: (maximum of 1 per component) This demo includes the properties, slots, and events tables, and allows for the user to modify properties in these table(s) which then affects the demo.
 
-Add the comment `<!-- docs: live demo -->` within the code block at the top and `$attributes` on the component tag. For example:
+Add the comment `<!-- docs: live demo -->` within the code block at the top and `$attributes` on the component tag. Note that if `defaults` is being used in order to set default demo values, the demo information pieces must be separated by newlines rather than be inline (e.g., `<!-- docs: live demo name:d2l-button -->`). For example:
 ```
 ```html
 <!-- docs: live demo
 name:<component-tag, e.g., d2l-button>
-size:<'small'|'medium'|'large'|'xlarge'>
+size:<'small'|'medium'|'large'|'xlarge'
+defaults:{"<attribute>": <value: "string"|boolean|number>}>
 -->
 <script type="module">
   import '@brightspace-ui/core/components/button/button.js';
--- a/docs/adding-component.md
+++ b/docs/adding-component.md
@@ -82,7 +82,7 @@ Add the comment `<!-- docs: live demo -->` within the code block at the top and 
 ```html
 <!-- docs: live demo
 name:<component-tag, e.g., d2l-button>
-size:<'small'|'medium'|'large'|'xlarge'
+size:<'small'|'medium'|'large'|'xlarge'>
 defaults:{"<attribute>": <value: "string"|boolean|number>}>
 -->
 <script type="module">
--- a/pages/assets/demo-snippet.js
+++ b/pages/assets/demo-snippet.js
@@ -1,6 +1,5 @@
 import '@brightspace-ui/core/components/colors/colors.js';
 import '@brightspace-ui/core/components/demo/demo-snippet.js';
-import './demo-tables.js';
 import './demo-resizable-preview.js';
 import 'playground-elements/playground-code-editor';
 import 'prismjs/prism.js';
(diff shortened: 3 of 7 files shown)


.eleventy.js
```diff
@@ -51,7 +51,7 @@ module.exports = function(eleventyConfig) {
 					let headingClass = token.attrGet('class') || '';
 					if (token.tag === 'h1') headingClass += ' d2l-heading-1';
 					else if (token.tag === 'h2') headingClass += ' d2l-heading-2';
-					else if (token.tag === 'h3') headingClass += ' d2l-heading-3';
+					else if (token.tag === 'h3') headingClass += ' d2l-heading-4';
```
margaree, line 54: I switched this to be styled like an h4 since that's what's happening in the `demo-tables` so this keeps that consistent.

pages/assets/demo-snippet.js
```diff
@@ -91,13 +91,13 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 				for (const attribute in this._attributes) {
 					const { type, value } = this._attributes[attribute];
 					switch (type) {
-						case 'String':
```
margaree, line 94: Switched this to be an object and use the same casing.

pages/assets/demo-snippet.js
```diff
@@ -115,22 +115,35 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
...
 		return name;
 	}
 
 	connectedCallback() {
 		super.connectedCallback();
+
+		if (this.defaults) {
+			const defaults = JSON.parse(this.defaults);
+			Object.keys(defaults).forEach((key) => {
+				const value = defaults[key];
```
margaree, line 143: I'm making an assumption that values will be sent as `"string"` or `14` (number) or `true / false` without quotes around the numbers or booleans and so we can pull type from that.

pages/assets/demo-snippet.js
```diff
@@ -171,30 +185,18 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
...
 		if (value === '' || !value) {
 			delete this._attributes[name];
 			this._attributes = { ...this._attributes };
 		} else {
 			this._attributes = { ...this._attributes, [name]: { type, value } };
 		}
-		this.requestUpdate();
-	}
-
-	_parseConfigurationValue(tag) {
```
margaree, line 184: This got moved to the utils file for testing.

pages/assets/demo-tables.js
```diff
@@ -49,10 +51,21 @@ export class ComponentCatalogDemoTables extends LitElement {
 		super();
 		this.hideSlots = false;
 	}
+
 	firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
 		this._componentInfo = components.find((component) =>  component.name === this.tagName);
+
+		this._componentInfo.attributes.sort((a, b) => {
```
margaree, line 60: I added a sort so that `required` attributes are first, following by alphabetical order, which is currently not guaranteed (I think order goes file properties, then mixin properties)

pages/assets/demo-tables.js
```diff
@@ -62,13 +75,23 @@ export class ComponentCatalogDemoTables extends LitElement {
...
 				demoType = info.type.replace(/\|/g, ' | ');
-				demoValue = this._getDemoValueOptions(demoType, info.name, infoDefault);
-			} else {
-				demoValue = validTypes.includes(demoType) ? this._getDemoValueOptions(demoType, info.name, infoDefault) : null;
 			}
+
+			if (this.interactive) {
+				if (Object.values(validTypes).includes(demoType)) {
+					const defaultForName = this._defaults[info.name];
+					if (defaultForName) initialValue = info.type !== 'string' ? `"${defaultForName}"` : defaultForName;
```
margaree, line 87: The current logic fails if the value is `true` and it isn't wrapped with `"` so I just wrapped everything with quotes.

pages/assets/demo-tables.js
```diff
@@ -80,14 +103,14 @@ export class ComponentCatalogDemoTables extends LitElement {
 				</tr>`;
 		});
 
-		const slotRows = this._componentInfo.slots.map((slotInfo) => {
+		const slotRows = this._componentInfo.slots ? this._componentInfo.slots.map((slotInfo) => {
```
margaree, line 106: There was a console error so I added this check.

pages/assets/utils.mjs
```diff
@@ -2,6 +2,28 @@ const defaultImports = [
 	'import \'@brightspace-ui/core/components/typography/typography.js?module\';\n',
 ];
 
+export function parseConfigurationValue(tag, demoSnippet, requireSplitOnNewlines) {
```
margaree, line 5: With `defaults`, we can't split on spaces and use inline since values can contain spaces (e.g., `"text": "My Button"`) so I made it necessary for if defaults is being used, this comment block must be split on newlines.

docs/adding-component.md
```diff
@@ -77,12 +77,13 @@ size:<'small'|'medium'|'large'|'xlarge'>
...
 **Interactive demo**: (maximum of 1 per component) This demo includes the properties, slots, and events tables, and allows for the user to modify properties in these table(s) which then affects the demo.
 
-Add the comment `<!-- docs: live demo -->` within the code block at the top and `$attributes` on the component tag. For example:
+Add the comment `<!-- docs: live demo -->` within the code block at the top and `$attributes` on the component tag. Note that if `defaults` is being used in order to set default demo values, the demo information pieces must be separated by newlines rather than be inline (e.g., `<!-- docs: live demo name:d2l-button -->`). For example:
 ```
 ```html
 <!-- docs: live demo
 name:<component-tag, e.g., d2l-button>
-size:<'small'|'medium'|'large'|'xlarge'>
+size:<'small'|'medium'|'large'|'xlarge'
```
GeorgePlukov, line 85: Missing the closing `>` on this line and the opening one on the defaults line

GeorgePlukov (changes requested):
Pushed up a commit to add an example to the demo.md page I've been using. I think there is something going on with the `Interactive demo - size` example, it seems to render the property table fine in master and with the defaults option on this branch, but with single line comments its failing.

GeorgePlukov (approved):

PR #183, opened 2021-07-01
Code parsing: move comment to before code block

margaree:
This allows for the code information comment to come before the code block.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -102,22 +102,25 @@ module.exports = function(eleventyConfig) {
 	const defaultHtmlRule = markdownIt.renderer.rules.html_block;
 	markdownIt.renderer.rules.html_block = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
-		if (content.includes('<!-- docs: live demo')) {
+		if (content.includes('<!-- docs: demo')) {
 			inCodeBlock = true;
-			const tag = parseConfigurationValue('name', content);
-			if (!tag) return '<d2l-component-catalog-demo-snippet resizable hide-code>';
+			let openingTag = '<d2l-component-catalog-demo-snippet resizable ';
 			const size = parseConfigurationValue('size', content);
-			const defaults = parseConfigurationValue('defaults', content, true);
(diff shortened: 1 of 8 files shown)


GeorgePlukov (approved):

.eleventy.js
```diff
@@ -117,7 +102,23 @@ module.exports = function(eleventyConfig) {
 	const defaultHtmlRule = markdownIt.renderer.rules.html_block;
 	markdownIt.renderer.rules.html_block = (tokens, idx, options, env, slf) => {
 		const content = tokens[idx].content;
-		if (content.includes('<!-- docs: start hidden content -->'))
+		if (content.includes('<!-- docs: live demo')) {
+			inCodeBlock = true;
+			const tag = parseConfigurationValue('name', content);
+			if (!tag) return '<d2l-component-catalog-demo-snippet resizable hide-code>';
+			const size = parseConfigurationValue('size', content);
```
GeorgePlukov, line 109: I was updating my branch with this and I _think_ the size value needs to be parsed outside of the if block and passed to any of the demo types (code demo/demo)

GeorgePlukov (changes requested):

.eleventy.js
margaree: Ah okay - previously I don't think we were supporting that based on the original if statement since it looked for the entire `<!-- docs: demo -->` etc but I can switch that since that would make sense to support.

margaree: Any objection if I change the comments to all start with `<!-- docs: demo` and then there'd be `<!-- docs: demo live`, `<!-- docs: demo code` and just `<!-- docs: demo` since that would make general parsing easier?

GeorgePlukov: Oh wow, I swear that I removed the closing fences from those at some point! And yeah that naming seems alright to me, I think those are referenced in the docs a couple places as well

margaree:
@GeorgePlukov This is ready for a re-review when you have a chance (no rush). If the `$attributes` related changes seem to be possible then I'll put up a separate PR for that.

GeorgePlukov (approved):

PR #184, opened 2021-07-01
Docs: Add page with instructions on GitHub edit workflow

margaree:


docs/editing-page.md
```diff
@@ -0,0 +1,19 @@
+# Editing a Page
+
+We use markdown files for our page content. These files are stored in GitHub. The GitHub UI workflow can be used to edit page content, with the steps shown below. Note that you will need a GitHub account and write access to the repository where you wish to make changes. For access, you can post in the #github Slack channel or reach out to your manager.
+
+To edit a page:
+1. At the bottom of each page in the Daylight site is text stating "Suggest an edit for this page" where "edit" is a link to the edit workflow. Click edit.
```
margaree, line 6: I think we were thinking this would be a temporary thing for component pages, but I thought it might be nice to leave this at the bottom of each page (included documentation repo ones for design docs etc) in order to make this workflow easier, rather than having instructions like "hunt through the repo in GitHub for the page you're looking for".

GeorgePlukov: I'm in favor of leaving an edit link as well. Maybe we make it a bit less prominent in the future.

docs/editing-page.md
```diff
@@ -0,0 +1,19 @@
...
+2. Edit will take you to a page that looks like the one below. Make the changes that you would like to make. You can then select the "Preview" tab near the top to preview your changes.
```
GeorgePlukov, line 7: Maybe its worth adding a note here that the preview won't look exactly like the published pages? I could see questions coming up about the page not having demos/formatting

GeorgePlukov (approved):

PR #195, opened 2021-07-07
Demo: remove $attributes

margaree:
This adds support for `allInstancesInteractive` in the code comment, which if true then all instances of the component under `name` are affected by properties table changes (and defaults) and if not present or false then only the first instance of that component is affected.

I also moved the contents of `code` out to a separate file so that I could add tests.

### DIFF
--- a/pages/assets/utils.mjs
+++ b/pages/assets/utils.mjs
@@ -38,12 +38,9 @@ export function getCode(snippet, interactive, currAttributes, tagName, allInstan
 	}
 	attributes.sort();
 	const attributesText = attributes.length === 0 ? '' : ` ${attributes.join(' ')}`;
-	const includesSpace = codeSnippet.includes(`<${tagName} `);
-	const replaceString = includesSpace ? `<${tagName}` : `<${tagName}>`;
-	const re = allInstancesInteractive ? new RegExp(replaceString, 'g') : replaceString;
-	return includesSpace
-		? codeSnippet.replace(re, `<${tagName}${attributesText}`)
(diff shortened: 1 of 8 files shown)


pages/assets/utils.mjs
```diff
@@ -2,6 +2,50 @@ const defaultImports = [
...
+			case validTypes.number:
+				attributes.push(`${attribute}=${value}`);
+				break;
+			default:
+				break;
+		}
+	}
+	attributes.sort();
+	const attributesText = attributes.length === 0 ? '' : ` ${attributes.join(' ')}`;
+	const includesSpace = codeSnippet.includes(`<${tagName} `);
```
GeorgePlukov, line 41: Does this support the case where there are some properties already existing on the element being edited. I think if the element has attributes predefined we would want them to stick around. Might not be something that we need to think about at the moment though

Basic example: `<d2l-button id="my-button"></d2l-button>`

margaree: It should. I'll confirm and then add a test for that as well.

margaree: Looks like it does and there is a test https://github.com/BrightspaceUI/documentation/pull/195/files#diff-d1c2db4e8e7f976d83970c5457c613a76c22c2e2d3a81ebb9336c68c584319b9R133 .

GeorgePlukov: Oh nice I didn't see that test on my first pass. Minor thing, we may want the interactive attributes to come after the predefined ones but looks good!

GeorgePlukov (approved):

PR #196, opened 2021-07-09
Button page: bump core

margaree:


GeorgePlukov:
This looks really nice!!!

dlockhart:
Wow, looking amazing!

Do we have a plan for when pages like this link to other components (e.g. button here is linking to Link and Dropdown) and maybe somehow being able to re-map that so that it actually takes you to the Daylight site's link/dropdown pages?

margaree:
My thinking was that after those components have pages in Daylight and the site is released, then we switch those links to point at the Daylight site ones.

It's a bit difficult to come up with an approach that would allow for relative links in the READMEs that also map correctly to the Daylight site. For something simple where the docs are in the same folder, something like `[my link](./other-page.md)` could be mapped to `./other-page.html` but for things in core, their links have to be either `../other-folder` or `../../components/other-folder` (assuming the files are all `README.md`) if we're referencing other components in other folders. If we did it like the latter example though I think there could be a pattern there with relative links that could work where we just ignore the `../..` and grab the `components/other-folder` piece (core only - hopefully this isn't as complicated in other components).

If we did that, then for a time we'd have some pages that don't exist (like currently link and dropdown) but it might work better long term and that's likely okay since in theory no one should be using the site yet.

margaree:
Closing in favour of https://github.com/BrightspaceUI/documentation/pull/198

PR #208, opened 2021-07-19
Automatic vertical resizing for demo components

GeorgePlukov:
Adds vertical resizing for demos based on the specified components height.

Notes:
- This is on by default, disable using `autoSize:false`
- Tag name must be provided (ex: `name:d2l-button`)


.eleventy.js
```diff
@@ -106,16 +106,23 @@ module.exports = function(eleventyConfig) {
 		if (content.includes('<!-- docs: demo')) {
 			inCodeBlock = true;
 			let openingTag = '<d2l-component-catalog-demo-snippet resizable ';
+
+			const autoSize = parseConfigurationValue('autoSize', content);
 			const size = parseConfigurationValue('size', content);
+			const tag = parseConfigurationValue('name', content);
+
 			if (size) openingTag += ` size="${size}" `;
+			if (tag && autoSize !== false) {
```
margaree, line 115: Are these parsed as strings? If so, should this be checking for `'false'`? (I'm not actually sure - could be useful to have a test in `eleventy-utils.test.js`)

nit: would be cleaner on one line like the previous line just for consistency

.eleventy.js
```diff
@@ -106,16 +106,23 @@ module.exports = function(eleventyConfig) {
...
+				openingTag += ' auto-size ';
+			}
 
 			if (content.includes('<!-- docs: demo live')) {
-				const tag = parseConfigurationValue('name', content);
+				const defaults = parseConfigurationValue('defaults', content, true);
```
margaree, line 120: Should defaults have moved here? If so, the usage should also move here, else it wouldn't be included in `openingTag`

pages/assets/demo-resizable-preview.js
```diff
@@ -41,7 +46,11 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			* @type {'small'|'medium'|'large'|'xlarge'}
 			*/
 			size: { type: String, reflect: true },
-			_previewWidth: { type: Number, reflect: true }
+			/**
```
margaree, line 49: nit: empty comment

pages/assets/demo-resizable-preview.js
```diff
@@ -106,27 +115,40 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			playground-preview {
 				border-radius: 10px;
 				height: 100%;
+				transition: height 0.2s ease;
```
margaree, line 118: We should probably also respect `prefers-reduced-motion` (e.g., https://github.com/BrightspaceUI/core/blob/main/components/calendar/calendar.js#L358)

pages/assets/demo-resizable-preview.js
```diff
@@ -106,27 +115,40 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
...
+		this._id = getUniqueId();
 	}
 	get indexHTML() {
 		return `
 			<script>
 				window.addEventListener('load', function () {
 					var demoEl = document.getElementById('demo-element');
 					demoEl.classList.remove('hide');
+					${this.autoSize && this.tagName ? `
+					window.requestAnimationFrame(function() {
```
margaree, line 148: nit: indentation

pages/assets/demo-resizable-preview.js
```diff
@@ -138,8 +160,11 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			<style>
 				/* todo?: add this to md template and provide configuration for different item alignments? */
 				html {
+					padding: ${DEMO_PADDING}rem;
+				}
+				html, body {
+					margin: 0;
```
margaree, line 166: nit: alphabetical order

pages/assets/demo-resizable-preview.js
```diff
@@ -169,6 +203,8 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 	render() {
 		const previewStyles = {
 			width: `calc(100% - ${this.resizable ? SLIDER_WIDTH : 0}px)`,
+			height: this._previewHeight ?  `calc(${this._previewHeight}px + ${ 2 * DEMO_PADDING}rem)` : undefined,
```
margaree, line 206: nit: alphabetical order and empty line

pages/assets/demo-resizable-preview.js
```diff
@@ -41,7 +46,11 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			* @type {'small'|'medium'|'large'|'xlarge'}
 			*/
 			size: { type: String, reflect: true },
-			_previewWidth: { type: Number, reflect: true }
+			/**
+			*/
+			tagName: { type: String, attribute: 'tag-name', },
+			_previewHeight: { type: Number, reflect: true },
```
margaree, line 52: Do these need to be reflected?

GeorgePlukov: Yeah when these are updated we need the component to cause a re-render since the styles are applied in the render method

.eleventy.js
GeorgePlukov: I was hoping to collect the `parseConfigurationValue` calls into one spot but never made it all the way I guess. I thought it would be nice to have all the argument parsing in one spot per code block but it would cause us to do some unnecessary work parsing every argument each time even if its not required. 

Might be something we want to look at doing if we start adding more configuration options

margaree (approved):
Nice!

PR #210, opened 2021-07-20
Page Generation Tweaks

margaree:
Changes I found were useful when preparing the input-date-time page


.eleventy.js
```diff
@@ -26,8 +27,11 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addShortcode('statusTable', (tier) => {
 		return `<d2l-component-catalog-status-table tier="${tier}"></d2l-component-catalog-status-table>`;
 	});
+	eleventyConfig.addFilter('cssmin', (code) => {
```
margaree, line 30: This was just moved

.eleventy.js
```diff
@@ -62,10 +66,16 @@ module.exports = function(eleventyConfig) {
 			}
 		}
 	};
-	const markdownIt = require('markdown-it')(options).use(require('markdown-it-modify-token'));
-	eleventyConfig.addFilter('cssmin', (code) => {
-		return new cleanCSS({}).minify(code).styles;
-	});
+	const anchorOpts = {
```
margaree, line 69: Adds a `d2l-icon` permalink similar to GitHub.

pages/_includes/css/style.css
```diff
@@ -81,6 +81,19 @@ h1, h2, h3 {
 .d2l-component-catalog-tag-inner {
 	color: var(--d2l-color-malachite-minus-1);
 }
+.header-anchor {
```
margaree, line 84: Styles for the heading links to hide then show on hover (can be tweaked in the future, this was just an easy place to start)

pages/assets/demo-resizable-preview.js
```diff
@@ -28,6 +28,10 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 			* Code for the preview IFrame to display
 			*/
 			code: { type: String },
+			/**
+			 * Corresponds to align-items value for demo content. Defaults to center.
+			 */
+			contentAlignment: { type: String, attribute: 'content-alignment' },
```
margaree, line 34: Since input-date-time etc open downwards, it became useful to be able to specify that they should be `flex-start` aligned

pages/assets/demo-tables.js
```diff
@@ -97,12 +97,14 @@ export class ComponentCatalogDemoTables extends LitElement {
 			}
 
 			const demoValueRow = this.interactive ? html`<td>${demoValue}</td>` : null;
+			const defaultShown = infoDefault ? infoDefault.replace(/'/g, '') : undefined;
```
margaree, line 100: These are more changes related to the new analyzer, where there were excessive `''`

tools/generate-pages.mjs
```diff
@@ -60,6 +60,8 @@ function _copyCustomElements(repos) {
 
 			const attributeMembers = declaration.members.filter((dec) => {
 				if (dec.name === 'properties' && dec.static) return false;
+				if (dec.name.startsWith('_')) return false;
```
margaree, line 63: Also new analyzer related - since it includes private properties that are prefaced with `_` this will filter those out for usage in Daylight.

pages/assets/demo-tables.js
```diff
@@ -181,6 +183,7 @@ export class ComponentCatalogDemoTables extends LitElement {
 						@change="${this._onNumberChange}"
 						data-name="${attributeName}"
 						label="${attributeName}"
+						label-hidden
```
margaree, line 186: We were showing the label for number input but hiding for text input so I added this to hide the label.

pages/assets/demo-tables.js
```diff
@@ -14,7 +14,7 @@ export class ComponentCatalogDemoTables extends LitElement {
 	static get properties() {
 		return {
 			defaults: { type: String },
-			hideSlots: { type: Boolean, attribute: 'hide-slots', reflect: true },
+			hideOtherTables: { type: Boolean, attribute: 'hide-other-tables', reflect: true },
```
margaree, line 17: @GeorgePlukov was `hide-slots` or the `d2l-slots-table` class being used?

margaree: Ah I see `hide-slots` in the tests. Will update (though a name other than `hideOtherTables` could be better, or we could have separate properties for hiding slots or events but not sure if there's a usecase for that?)

GeorgePlukov: `hide-slots` is just there in case there is a case where we want to prevent the interactive demos from rendering slots. The `d2l-slots-table` class was added so that we can easily determine where the table is being generated from

.eleventy.js
GeorgePlukov: The link buttons look great!

GeorgePlukov (approved):

PR #213, opened 2021-07-21
Tweaks found when preparing Menu page

margaree:


.eleventy.js
```diff
@@ -122,7 +122,7 @@ module.exports = function(eleventyConfig) {
 			const tag = parseConfigurationValue('name', content);
 
 			if (size) openingTag += ` size="${size}" `;
-
+			if (tag) openingTag += ` tag-name="${tag}" `;
```
margaree, line 125: I moved this here so the regular, non-interactive demo (e.g., the one at the top of a page) could also auto-resize.

pages/assets/demo-resizable-preview.js
```diff
@@ -158,6 +158,7 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 					${this.autoSize && this.tagName ? `
 						window.setTimeout(function() {
 							var demoSizeElement = document.querySelector('${this.tagName}');
+							if (!demoSizeElement) return;
```
margaree, line 161: The `getBoundingClientRect` was throwing if the element wasn't found.

.eleventy.js
GeorgePlukov: I had this exact change locally atm! Noticed it while I was prepping the switch page

pages/assets/demo-resizable-preview.js
GeorgePlukov: For autosizing to work you have to set a component name/tag. We could change it to use its own selector potentially 🤔

GeorgePlukov (approved):

pages/assets/demo-resizable-preview.js
margaree: Looks like we can use `var demoSizeElement = demoEl.querySelector(':not(script)');` to grab the first non-script element and that would work and then we could not need to rely on tag.. Any concerns around this approach? (other than we'd just have to be watchful that there aren't demos doing something unexpected with the first element). This would actually be helpful with menu since I have demos where the tag is `d2l-menu-item` so I can't use autosize since it sizes to the menu item and not the outer `menu`.

GeorgePlukov: Thats seems like it would work reasonably well. Im not sure if we want to be using the autosizing on menus though

margaree: It works pretty well with the menu demos. We might want a set size on the nested ones, though those still work with scrolling, since the height changes.

pages/assets/demo-resizable-preview.js
```diff
@@ -155,11 +151,12 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
...
 					demoEl.classList.remove('hide');
-					${this.autoSize && this.tagName ? `
+					${this.autoSize ? `
 						window.setTimeout(function() {
-							var demoSizeElement = document.querySelector('${this.tagName}');
+							var demoSizeElement = demoEl.querySelector(':not(script)');
+							if (!demoSizeElement) return;
 							window.parent.postMessage({id: "${this._id}", height: demoSizeElement.getBoundingClientRect().height}, '*');
-						}, 20)` : '' }
+						}, 100)` : '' }
```
margaree, line 159: I noticed that sometimes the height wasn't autosizing and that `getBoundingClientRect().height` would be returning 0. I extended the timeout to hopefully minimize the amount this happens.

GeorgePlukov: Switched this to using a timeout recently, it was using `requestAnimationFrame` but I was seeing this happen. Increasing the timeout should work but maybe there is a better solution we could implement? 

One possibility would be to start polling every X ms once the page loads until we get a non-zero height. That way it would load a bit more quickly than 100ms and still get the height if for some reason it is > 100. I think we can tackle this at the same time as when we add intersection observer to only load demos that are currently in view

margaree: Yeah I think I like the polling idea better. I'll make that change.

pages/assets/demo-resizable-preview.js
```diff
@@ -155,11 +151,24 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
...
+						var demoSizeElement = demoEl.querySelector(':not(script)');
+						if (!demoSizeElement) return;
+
+						var numRepeats = 0;
+
+						var setHeight = function() {
+							var height = demoSizeElement.getBoundingClientRect().height;
+							if (height !== 0) {
+								window.parent.postMessage({id: "${this._id}", height: demoSizeElement.getBoundingClientRect().height}, '*');
+							} else if (numRepeats < 5) {
```
margaree, line 164: I set this to 5 max repeats just so it doesn't keep looping forever if for some reason it's never getting a height. If we're still seeing this issue then we can try bumping this value.

PR #215, opened 2021-07-21
Parse code for attributes, remove defaults and allInstancesInteractive

margaree:
As mentioned here https://github.com/BrightspaceUI/core/pull/1503#discussion_r674049071, it's a bit unfortunate that our examples aren't able to include attributes. This change parses them out of the code rather than using defaults. I removed defaults and `allInstancesInteractive` because it becomes harder to support/confusing this way, and I don't know that there's really a great case for using it anyways.


package.json
```diff
@@ -10,10 +10,11 @@
     "lint:eslint": "eslint . --ext .js,.mjs",
     "lint:lit": "lit-analyzer pages/**/*.js --strict",
     "lint:style": "stylelint \"**/*.js\"",
-    "start": "cross-env NODE_ENV=development npm run build && eleventy && concurrently \"npx eleventy --watch\" \"web-dev-server\"",
+    "start": "cross-env NODE_ENV=development npm run build && eleventy && npm run start:servers",
+    "start:servers": "concurrently \"npx eleventy --watch\" \"web-dev-server\"",
     "start:prod": "npm run build:prod && web-dev-server --config web-dev-server-prod.config.js",
     "test": "npm run lint && npm run test:unit",
-    "test:unit": "mocha './test/**/*.test*js'"
+    "test:unit": "mocha './test/**/*.test*js' -r jsdom-global/register"
```
margaree, line 17: So this is so we can use `DOMParser` and run mocha tests. I don't know if there's a better way?

pages/assets/utils.mjs
```diff
@@ -38,11 +38,30 @@ export function getCode(snippet, interactive, currAttributes, tagName, allInstan
 	}
 	attributes.sort();
 	const attributesText = attributes.length === 0 ? '' : ` ${attributes.join(' ')}`;
-	const reString = `(<${tagName})(>)|(<${tagName} [^>]+)(>)`;
-	const re = allInstancesInteractive ? new RegExp(reString, 'g') : new RegExp(reString);
+	const re = new RegExp(`(<${tagName})(>)|(<${tagName} [^>]+)(>)`);
 	return codeSnippet.replace(re, `$1$3${attributesText}>`);
 }
 
+export function getDefaults(code, tagName) {
```
dlockhart, line 45: Instead of regexes, would it be possible to just create the element using a `<template>` and then iterate over its attributes? Something like:

```javascript
const t = document.createElement('template');
t.innerHTML = code;
const defaults = {};
[...t.content.firstChild.attributes].forEach((attr) => {
...
});
```

margaree: Yup that works!

package.json
GeorgePlukov: Seems like this could live in a `mocha.opts` file but I'm fine with it being here until we add more configuration maybe

GeorgePlukov (approved):

PR #220, opened 2021-07-26
Adding component docs page improvements

margaree:


.eleventy.js
```diff
@@ -115,12 +115,12 @@ module.exports = function(eleventyConfig) {
...
 			let openingTag = '<d2l-component-catalog-demo-snippet resizable ';
 
-			const autoSize = parseConfigurationValue('autoSize', content);
-			if (autoSize !== 'false') openingTag += ' auto-size ';
-
 			const size = parseConfigurationValue('size', content);
 			if (size) openingTag += ` size="${size}" `;
 
+			const autoSize = parseConfigurationValue('autoSize', content);
+			if (autoSize !== 'false' && !size) openingTag += ' auto-size ';
```
margaree, line 122: I thought it made sense that if size is set, then the demo should be that size instead of auto-sized. Currently it seems that `auto-size` takes priority.

GeorgePlukov: The way I imagined this working was that the size property would set the initial size of the demo to resize from. That way you could kinda estimate which size would be closest and once the demo loads it would make a smaller adjustment to the container size.

margaree: Ok I'll put that back to how it was before then.

GeorgePlukov (approved):

PR #249, opened 2021-07-28
demo-tables fixes

margaree:
- default value for specified strings
~- support object as string (in tooltip, boundary is an object which is initially inputted as a string, so using a text input for this seems to work)~
- handle value being 0


GeorgePlukov:
I remember us discussing that we didn't want to display the boundary item in the interactive demos. Would this just allow you to type an object as a string into the text field?

margaree:
It would, yeah. We can also just leave it out since there is the section about boundaries.

GeorgePlukov (approved):

PR #251, opened 2021-07-28
Adds display property for demo configurations

GeorgePlukov:


pages/assets/demo-resizable-preview.js
```diff
@@ -218,7 +223,7 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 				}
 				.layout {
 					align-items: ${align};
-					display: flex;
+					display: ${display};
 					flex-wrap: wrap;
```
margaree, line 227: I guess it'd probably be a bit messy to not have the `flex` related properties unless display is set to flex.

margaree (approved):
Can you also add this to the documentation https://github.com/BrightspaceUI/documentation/blob/main/docs/adding-component.md#code-blocks ?

pages/assets/demo-resizable-preview.js
GeorgePlukov: Should be able to do this without making it too confusing, Ive got something in the works

margaree (approved):
Looks good!

PR #252, opened 2021-07-28
Design improvements to demo component

GeorgePlukov:
Demo updates according to design:
- Min width from 300 -> 320px
- Slider width from 35 -> 25px
- Cursor to ew-resize
- increases snap open size to 30px (can reconfigure after playing with it for a bit)
- Adds size text when repositioning slider
- Hover color changes
- Fixes issue with dark mode causing normal slider to be slightly too high (rework coming for this)


pages/assets/demo-resizable-preview.js
```diff
@@ -341,7 +362,7 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 
 		const updatedWidth = Math.max(MINIMUM_WIDTH, this._previewWidth - intervalWidth);
 		this._previewWidth = Math.round(updatedWidth / intervalWidth) * intervalWidth;
-
+		this._fullyOpen = false;
```
GeorgePlukov, line 365: Open to different naming options on this variable

pages/assets/demo-resizable-preview.js
```diff
@@ -292,21 +311,23 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
...
-							aria-label="Resizable demo slider"
-							aria-describedby="instructions"
-							aria-orientation="vertical">
-							<svg class="d2l-slider-icon" width="5" height="18" viewBox="0 0 5 18" xmlns="http://www.w3.org/2000/svg">
-								<g  fill-rule="evenodd">
-									<rect width="2" height="18" rx="1"/>
-									<rect x="3" width="2" height="18" rx="1"/>
-								</g>
-							</svg>
+						<div class="d2l-slider-container" style=${styleMap(sliderStyles)}>
```
GeorgePlukov, line 314: Moved the dynamic positioning to a container element as it made it a bit easier to add the `${pixels} px` text

margaree (approved):

pages/assets/demo-resizable-preview.js
```diff
@@ -171,6 +182,7 @@ class ComponentCatalogDemoResizablePreview extends LitElement {
 		this.darkMode = false;
 		this.size = 'small';
 
+		this._fullyOpen = true;
```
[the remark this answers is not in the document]
GeorgePlukov: I was thinking it would be ideal to always show the pixel value when you were dragging the slider. While I was testing it seems to still disappear after dropping the slider. I don't think adding _fullyOpen as a property would fix this but when I review the design with Jeff I can make sure this is working properly. Maybe when there is keyboard focus we always show pixel value?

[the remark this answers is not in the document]
margaree: Sounds good, yeah just run it by Jeff to see if it's all as expected. Otherwise this LGTM.

package-lock.json
```diff
@@ -68,9 +68,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.14.5",
```
margaree, line 71: Did you mean to commit this file?

margaree (commented):
Looks pretty good!

I noticed the following when testing:
- the time for the px value to disappear post-hover/focus is quite a while (though that might be intended)
- dark mode doesn't have any focus/hover style (totally fine if this is being done separately)
- if I resize the demo to be smaller than max, then resize the page to be less than the demos width, then hover on the slider, the px value doesn't update (hopefully that all made sense)

GeorgePlukov:
> * the time for the px value to disappear post-hover/focus is quite a while (though that might be intended)

I lowered the time it stays on screen to 0.5s from 1s, seems a lot better now

> * dark mode doesn't have any focus/hover style (totally fine if this is being done separately)
Changes are in a separate PR 

> * if I resize the demo to be smaller than max, then resize the page to be less than the demos width, then hover on the slider, the px value doesn't update (hopefully that all made sense)

Nice catch! Fixed this by moving the pixel value calculation into the `updated` lifecycle method. Thanks for testing it out

margaree (approved):
One thing I noticed on re-test is that when I focus on the slider with keyboard but then use the mouse to move the slider, it maintains the keyboard focus style. Not sure if this is expected.

PR #255, opened 2021-07-30
demo tables: support markdown parsing in description

margaree:
When prepping the numeric input pages, there were some descriptions that would ideally contain code and links, so this allows those to work here, rather than just showing up as string containing the markdown attributes like [ my link ] ( www.link.com ) . There are others we could handle as well, this was just the first ones I've noticed we've needed for these descriptions.


pages/assets/demo-tables.js
```diff
@@ -273,5 +281,10 @@ export class ComponentCatalogDemoTables extends LitElement {
 		this._dispatchChangeEvent(eventDetails);
 	}
 
+	_parseMarkdown(text) {
+		if (!text) return;
+		return text.replace(RE_CODE, '<code>$1</code>').replace(RE_LINK, '<d2l-link href="$2">$1</d2l-link>');
```
GeorgePlukov, line 286: Seems like this is a fairly simple way to get this done. Do you imagine us needing to support more features from markdown within these descriptions? I'm thinking it may be easier to use a slightly more legit markdown parser here to make it easier to add support for other items in the future

margaree: Yup that's possible. I'll see if we can just pull in `markdown-it` that we're already using with 11ty

margaree: I wasn't able to get `markdown-it` to work (couldn't figure out how to get it to import) but was able to get `marked` working.

GeorgePlukov (approved):

PR #283, opened 2021-08-11
Add update-package-lock workflow

svanherk:
Left off auto-approve (which requires changes to the CODEOWNERS file and is most useful with auto-merge) and auto-merge for now.  Guessed at a schedule but let me know if you have other preferences - we can always change it if it's too noisy or not noisy enough.  I also left all the commit message / branch name/ pr title defaults but we can customize those if desired: https://github.com/BrightspaceUI/actions/tree/master/update-package-lock#using-the-action


GeorgePlukov (approved):
LGTM. That schedule seems like it should be fine for now, might reduce the amount of days if it gets really noisy.

PR #293, opened 2021-08-17
dependabot: ignore patch updates

margaree:
Based on https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#specifying-dependencies-and-versions-to-ignore this should hopefully ignore the patch updates.


GeorgePlukov (approved):
Nice find!

PR #320, opened 2021-09-17
New Style section with Color page

geurts:
Pushing what I have for now. Apparently eleventy doesn't like spaces in the folder names? Took me a bit to figure that out. 

I'm not sure what's up with the code block near the top of the Colors page, it's not rendering as a code block.


GeorgePlukov:
I added a new template to import the JS that runs the code demo so it should be appearing now! The page looks really great! 

Also for future PR's we typically try and avoid commit the `package-lock.json` file since the dependencies can change slightly based on the machine. I reset it on this pr, but that's why the build was failing for the pr deploy.

margaree:
There looks to already be a `demo` layout that I think is doing the same thing (just with more copy/pasted from the base layout). Would we want to either use that one or switch this new one to have that name?

GeorgePlukov:
> There looks to already be a `demo` layout that I think is doing the same thing (just with more copy/pasted from the base layout). Would we want to either use that one or switch this new one to have that name?

Makes sense, I renamed the demo layout to `basic.njk`, its just the base + demo-snippet

GeorgePlukov (approved):

PR #322, opened 2021-09-17
Hidden content in code blocks

GeorgePlukov:


pages/assets/demo-snippet.js
```diff
@@ -112,20 +115,24 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 		return parseImports(this._demoSnippet);
 	}
 
+	get strippedCode() {
```
GeorgePlukov, line 118: Not super happy with this name

pages/assets/demo-snippet.js
```diff
@@ -10,6 +10,9 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
 import { themeStyles } from './code-style.js';
 import { unsafeHTML } from 'lit-html/directives/unsafe-html';
 
+// eslint-disable-next-line no-control-regex
```
GeorgePlukov, line 13: This rule is disabled to allow us to search for hidden/escaped characters in this case the option `\n` that we want to support

GeorgePlukov:
Taking a look at the failing tests

pages/assets/demo-snippet.js
margaree: Hmm.. demoCode? preferredCode?

margaree (approved):

PR #342, opened 2021-09-29
HTML Editor related changes

margaree:
1. Made generation of `imports` for `importmap` a bit cleaner/easier to add to
2. There was an issue with parsing pages containing bash/shell code, so I added an import along with a check and backup for prism having the language


pages/assets/demo-resizable-preview.js
```diff
@@ -18,6 +18,17 @@ const SLIDER_WIDTH = 25;
...
 const STEP_COUNT = 6;// Number of steps that keyboards have while resizing
 
+const repoImports = [
+	'@brightspace-ui/core',
+	'@brightspace-ui/htmleditor',
+	'lit-element'
+];
+const IMPORTS = {};
+repoImports.forEach((repo) => {
+	IMPORTS[repo] = `https://unpkg.com/${repo}@latest`;
```
GeorgePlukov, line 28: I think we may want to be able to configure versions here still, when I was working on the `html-block` page I attempted to import `unsafeHtml` but in version `3.xx` the import had moved. Did we want to keep the `latest` for core as well? I was under the impression that we would have it at the latest for now but once we launch it would get locked to the same version the site is using (otherwise we could have mismatched demo code and attributes/events in our tables)

pages/assets/demo-snippet.js
```diff
@@ -1,6 +1,7 @@
 import './demo-resizable-preview.js';
 import './demo-tables.js';
 import 'prismjs/prism.js';
+import 'prismjs/components/prism-bash.min.js';
```
GeorgePlukov, line 4: Sweet

pages/assets/demo-resizable-preview.js
margaree: Agreed - I think we'd want the site version and demo version to be the same. I can work on that as part of this PR.

GeorgePlukov (approved):

margaree:
Closing for now - there's more changes to make so I'm just going to open a new PR.

PR #345, opened 2021-10-04
HTML Editor related changes

margaree:
~This can be ignored for now.~

This PR does the following:
- Adds a `disable-iframe` property to demo-snippet that will allow the demo to be shown without being in the iframe.
- Sets demo version to be the same as page version


pages/assets/demo-snippet.js
```diff
@@ -181,7 +202,13 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 
 			}
 			this._demoSnippet = code;
-			this._highlightedCodeSnippet = Prism.highlight(this.strippedCode, Prism.languages[this.language], this.language);
+			this._highlightedCodeSnippet = Prism.highlight(this.strippedCode, this._getPrismGrammar(), this.language);
+			if (this.disableIframe) {
```
margaree, line 206: This is hard coded for now because this will be used as a prototype and this was the most complex piece to implement. If this workflow seems good to go then I'll update this to work better. Also I'll fix the `node_modules` piece to point at /assets

GeorgePlukov: Will this always be doing `html-editor` specific things?  if it is maybe we just change `disableIframe` to something that makes that more obvious.

margaree: I think it's possible we encounter these problems in other cases (e.g., components where lang terms are dynamically imported in this way such that unpkg doesn't handle the case) so if we keep it generic we'd be saving future us from having to make it generic. I'll work on this today so I'll see if it ends up being straightforward to do this or not. If it's super complex then I might leave it as html editor specific for now and rename `disableIframe` and then create a story for MVP to improve this.

.eleventy.js
```diff
@@ -17,6 +19,17 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addNunjucksShortcode('issue', (issueUrl) => {
 		return `Looking for more details on the component or want to add your input? <d2l-link href="${issueUrl}">Check out the GitHub Issue</d2l-link>`;
 	});
+	eleventyConfig.addNunjucksShortcode('imports', (components) => {
```
margaree, line 22: This was added in order to add the script to the page to import the components.

Basically we want for dev for the demo/page to import the component from node_modules and in prod it needs to import from _site/assets. It would likely also be possible to do this from demo-snippet, but there would also be some changes needed in order to know if we're in prod and to read the manifest, and we'd still need to know the components that need to be rolled up, so this approach seemed somewhat more straightforward.

pages/assets/demo-resizable-preview.js
```diff
@@ -18,6 +19,13 @@ const SLIDER_WIDTH = 25;
 const MINIMUM_WIDTH = 320 + SLIDER_WIDTH;
 const STEP_COUNT = 6;// Number of steps that keyboards have while resizing
 
+componentVersions['lit-element'] = 'latest';
```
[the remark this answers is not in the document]
margaree: Forgot to fix this - will fix

pages/assets/demo-snippet.js
```diff
@@ -74,10 +79,15 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 				display: none;
 			}
 
-			:host([code-only]) pre {
+			:host([code-only]) pre,
+			:host([disable-iframe]) pre {
 				border-radius: 10px;
 			}
 
+			.d2l-demo-snippet-content {
```
margaree, line 87: Will also take a look at the fixed height bit

pages/assets/utils.mjs
```diff
@@ -58,25 +58,28 @@ export function getDefaults(code, tagName) {
 	return defaults;
 }
 
-export const parseImports = (allContent) => {
+export const parseScripts = (allContent, removeImports) => {
```
margaree, line 61: This is in order to keep any other functionality of the script other than the imports, which would be done at page level

rollup.config.js
```diff
@@ -19,17 +22,26 @@ export default [
...
 		plugins: [
 			outputManifest({
 				fileName: '../../pages/_includes/manifest.json',
 				publicPath: 'assets/'
 			}),
+			copy({
+				targets: [{
+					src: './node_modules/@brightspace-ui/htmleditor/tinymce/**',
+					dest: '_site/assets',
+					rename: (name, extension) => `../../../../tinymce/${name}${extension ? `.${extension}` : ''}`
```
margaree, line 36: Let me know if there's a better/less hacky way to do this - it copies the entire nested directory structure

pages/assets/demo-snippet.js
```diff
@@ -181,7 +207,37 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
...
+
+			if (this.disableIframe) {
+				const myScript = document.createElement('script');
+				myScript.setAttribute('type', 'module');
+				myScript.textContent = this.scripts;
+				document.querySelector('head').appendChild(myScript);
+
+				let numRepeats = 0;
+
+				const setHeight = () => {
```
margaree, line 220: This sets the height of the component div wrapper to be fixed to whatever the initial height is. The pro of this is there isn't document reflow when a property changes. The con is that if height is somehow made to be larger than this height, that wont be reflected. I can add an overflow scroll to this so that at least it would all be visible?

pages/assets/demo-snippet.js
```diff
@@ -181,7 +207,37 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
...
+						return;
+					}
+
+					const demoSizeElement = demoEl.querySelector(':not(script):not(style)');
+					const height = demoSizeElement.getBoundingClientRect().height;
+					if (height !== 0) {
+						this._demoHeight = height;
+					} else if (numRepeats < 6) {
+						numRepeats++;
+						setTimeout(() => setHeight(), 200);
```
margaree, line 235: I found this worked better with a longer timeout. Could also decrease timeout and increase numRepeats

tools/generate-pages.mjs
```diff
@@ -85,6 +89,7 @@ function _copyCustomElements(repos) {
 	_writeJSONToGeneratedFile(tags, FILENAME_CUSTOM_ELEM);
 }
 function _copyMarkdown(files) {
+	fs.mkdirSync(DIR_IMPORTED_COMPONENTS, { recursive: true });
```
margaree, line 92: I added this mkdirs in because if these functions got rearranged and one was relying on a different function having created a directory then the run could fail, so this makes them more self-reliant.

pages/assets/demo-snippet.js
GeorgePlukov: Seems like we are doing something pretty similar in our resizable preview component with the height setting. Does it make more sense for the disableIframe option to live within there? Then we would at least be dealing with anything preview related within the same file.

tools/generate-pages.mjs
```diff
@@ -142,14 +150,24 @@ function _getInfoGeneratePage(issue) {
...
 	let dependencyString = '';
 	return new Promise((resolve) => {
-		dependencies.forEach((dep) => {
-			exec(`npm list ${dep}`, (error) => {
+		repoInstallLocations.forEach((dep) => {
+			exec(`npm list ${dep} --json --depth 0`, (error, res) => {
 				if (error) dependencyString += ` ${dep}`; // dependency is not installed
+				else {
+					const resJSON = JSON.parse(res);
+					if (!resJSON.dependencies || !resJSON.dependencies[dep]) dependencyString += ` ${dep}`; // dependency is not installed
```
GeorgePlukov, line 168: nit: I'm personally not a fan of removing the curly braces from if statements in spots with nested logic

pages/assets/demo-snippet.js
margaree: So I considered that but the resizable preview component is so specific to supporting the playground pieces that it seemed not great to put this piece in there that would basically not make use of any of what's there. It might be possible to pull setHeight out into a shared function though.

pages/assets/demo-resizable-preview.js
```diff
@@ -272,21 +290,10 @@ class ComponentCatalogDemoResizablePreview extends FocusVisiblePolyfillMixin(Lit
...
-								window.parent.postMessage({id: "${this._id}", height: demoSizeElement.getBoundingClientRect().height}, '*');
-							} else if (numRepeats < 5) {
-								numRepeats++;
-								window.setTimeout(function() {
-									setHeight();
-								}, 20);
-							}
-						};
-						setHeight();` : '' }
+						var setHeight = ${setHeight};
```
GeorgePlukov, line 293: Kinda cool that this works

PR #364, opened 2021-10-14
initial perf experiments

margaree:


pages/assets/utils.mjs
```diff
@@ -83,3 +81,22 @@ export const parseScripts = (allContent, removeImports) => {
...
+	if (matches.length !== 3) throw new Error(`Error processing import "${line}"`);
+	const baseName = matches[2]; // "component.js"
+	const pre = matches[1]; // "import [... from] "
+
+	const url = window.location;
+	const baseUrl = `${url.protocol}//${url.host}`;
+
+	let result = '';
+	if (manifest[baseName]) {
+		result = `${pre} '${baseUrl}/${manifest[baseName]}';`;
```
GeorgePlukov, line 97: Instead of replacing the imports with the urls from the manifest is it possible to change the `unpkg` link in the import map to just point at this new endpoint?  Not sure if that would work as expected but i think it was the intent behind the import map

.github/workflows/deploy.yml
```diff
@@ -42,6 +42,11 @@ jobs:
         env:
           # cred variables set in the "Assume role" step
           AWS_DEFAULT_REGION: ca-central-1
+      - name: Deploy - playground elements
+        run: aws s3 sync --delete ./playground-build/ "s3://daylight.d2l.dev/branches/playground-build"
```
margaree, line 46: This looks to be different from https://github.com/Brightspace/d2l.dev/blob/master/sites/daylight/routes.yml#L12

GeorgePlukov (approved):
LGTM

GeorgePlukov:
Hmm, Just noticed that Icons arent working properly

PR #368, opened 2021-10-15
Pilot: Banner initial work, reorganize layouts

margaree:
- Added a dismissible `d2l-alert` to top of the page
- The link doesn't have an href yet - we'll add this once the Confluence page actually exists
- I removed `basic` layout to avoid duplication, then created a `demo` which builds off `base` layout and adds the demo snippet import


GeorgePlukov:
If you wanted could also add a slight animation to the page content so that it slides up

PR #369, opened 2021-10-15
Demo: Non-iframed: Wrap getting height in a timeout to allow time for child to completely render

margaree:
The demo was getting a demoHeight of 27 since the label was initially loaded, and then since it doesn't change the height, since the goal of this piece is to have a consistent height when properties get changed, it would be stuck at 27px when the full html editor loaded.

I tried many combinations of `await updateComplete`s and `requestAnimationFrame`s, but `setTimeout` was the only thing I could get working that was generic enough.

This whole piece may change in the future (e.g., maybe we do want a start height that we resize on first load that's larger than that, or something like that) so this is hopefully just a bit of a bandaid for the current break.


margaree:
Since this is now broken in the main Daylight deployment, I'm just going to merge this one and we can figure out an alternate solution if wanted in the future.

PR #384, opened 2021-10-26
Tweaking pilot banner appearance/language

geurts:


pages/_includes/layouts/base.njk
```diff
@@ -12,6 +12,22 @@
 	</style>
 	<script src="{{assets.getPath("base-imports.js")}}" type="module"></script>
 	<script src="{{assets.getPath("navigation.js")}}" type="module"></script>
+	<script>
+		let showPilotHeader;
+
+		window.addEventListener('DOMContentLoaded', function() {
+			showPilotHeader = window.localStorage.getItem('hide-pilot-alert');
```
margaree, line 19: nit: Probably would want to call this `hidePilotAlert`

pages/_includes/layouts/base.njk
```diff
@@ -12,6 +12,22 @@
...
+	
+			if (showPilotHeader === 'true') {
```
margaree, line 21: I feel like we might want to do the reverse of this, where we hide by default and then show if `hidePilotAlert` is `false`, since otherwise we do get the slide in animation on pages after the banner has been closed. Otherwise this LGTM!

PR #395, opened 2021-10-29
Update Daylight look

margaree:
Initial pass to get Daylight looking more like the designs. This can be tweaked if the idea for the h2s changes but gets the basic work in place.

Doesn't handle ~hover~, animations, or h2 styling scrolled to/clicked on yet.

Also the nav title piece does a bit of an ugly FOUC that I hope to fix separately.


pages/about/overview.md
```diff
@@ -0,0 +1,22 @@
...
+  parent: about
+---
+
+# Overview
+
+Daylight is a living design system that unites all of D2L's products across devices and market segments to ensure a consistent user experience - at scale!
+
+This website is where you'll find the information, specifications, and resources to design and build great products with Daylight.
+
+## Getting Started
```
margaree, line 16: Left this empty for now since we don't have Design and Developer pages yet

pages/assets/navigation.js
```diff
@@ -114,6 +123,9 @@ class DaylightNavigation extends LitElement {
 					else return 0;
 				});
 			}
+			if (!entry.parent || !entry.parent.parent) {
```
margaree, line 126: This is to open the first child page for the top and secondary pages with children

pages/style-elements/illustration.md
```diff
@@ -7,7 +7,7 @@ eleventyNavigation:
 ---
 
 <style>
-  img {
+  img.d2l-daylight-illustrations-img {
```
margaree, line 10: These were overriding other img rules so I made them more specific

pages/components/status.md
```diff
@@ -9,9 +9,9 @@ eleventyNavigation:
 
 # Component Status
 
-This page lists all of our components, including [Labs]() components and [requested]() components. We include the overall status of each component, but for more details follow the link on the component name.
+This page lists all of our components, including [Labs](#labs-components) components and [requested](#requested-components) components. We include the overall status of each component, but for more details follow the link on the component name.
```
margaree, line 12: I noticed these weren't real links so I updated them and temporarily linked the `Component Tiers` to the guide

pages/_includes/css/style.css
```diff
@@ -43,33 +43,41 @@ d2l-alert.d2l-component-catalog-warning-banner {
...
 .d2l-component-catalog-title {
 	border-bottom: 1px solid #e6eaf0;
-	display: flex;
-	justify-content: center;
-	padding-bottom: 0.75rem;
-	padding-top: 0.75rem;
+	padding: 1.15rem 0 0.5rem 1.25rem;
+}
+.d2l-component-catalog-title div.d2l-heading-3 {
+	margin-bottom: 0 !important;
```
GeorgePlukov, line 62: Is there no way to set these using more specific css selectors?

GeorgePlukov (approved):

pages/_includes/css/style.css
margaree: Actually yeah this seems to already be specific enough. Will remove

PR #403, opened 2021-11-02
About: "What's New" page

margaree:
Not totally loving the `Design Site Development` heading. Also thoughts on if content should be in table with date and change columns, or just a list?


pages/about/whats-new.md
```diff
@@ -0,0 +1,16 @@
...
+  parent: about
+---
+
+# What's New
+
+This page will stay up-to-date with changes to components (e.g., new component pages) and to the design site iteself (e.g., new/changed functionality).
+
+## Components
+
+## Design Site Development
```
GeorgePlukov, line 16: Maybe we change this one to just be the url? `daylight.d2l.dev`?

GeorgePlukov:
Sorry missed the question about formatting. I feel like just using text should be fine for now, at least for our documentation site. I think this is a pretty classic format to follow https://keepachangelog.com/en/1.0.0/

Not sure how we should structure the component updates. Honestly if we wanted to track actual updates to components/pages I think it should happen within that components README and then pulled in as part of the build process. It would be a pretty dramatic change to introduce a changelog for every component though. Might be an issue we bring up with the team?

margaree:
Yeah for components at least initially I was thinking we just included components that have pages added, which we can update as a part of that "Published" auto-generated PR. If we'd want more detailed notes, we'd likely have to go into release notes which could get complex, and yeah something like pairing that with the README.

pages/about/whats-new.md
```diff
@@ -0,0 +1,24 @@
...
+  parent: about
+---
+
+# What's New
+
+This page will stay up-to-date with changes to components (e.g., new component pages) and to the design site iteself (e.g., new/changed functionality). The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
+
+## Components
+
+### YYYY-MM-DD
```
margaree, line 16: I added these in as an example with the idea we would replace them once there are actual changes

GeorgePlukov (approved):

PR #409, opened 2021-11-03
Filter "Internal event", move sort into generate-pages script

margaree:
The "Internal event" piece was was the Selection page. I thought I would move the sort while I was there since it doesn't really need to be done in `demo-tables` and can be done in advance


tools/generate-pages.mjs
```diff
@@ -50,44 +50,88 @@ function _copyCustomElements(repos) {
...
-
-				if (attributeMember.description || !declaration.attributes) return;
-
-				const attributeName = attributeMember.attribute;
-				const attributeData = declaration.attributes.find((attribute) => attribute.name === attributeName);
-				if (attributeData) attributeMember.description = attributeData.description;
+		} else if (parsed.tags) {
+			parsed.tags.forEach((component) => {
+				if (component.attributes) {
+					component.attributes.sort((a, b) => {
```
GeorgePlukov, line 111: Maybe pull this sort out in a helper function if they seem to be the same/similar

GeorgePlukov (approved):

PR #413, opened 2021-11-05
Initial work on full component page visual-diffs

margaree:
This is the initial work on this process. Currently it only runs for `alert` and `button` and does not display the demos. If this format seems good, I'll add running for other pages (can be a part of this PR) and then work on displaying the demos in a subsequent PR.


.github/workflows/ci.yml
```diff
@@ -28,6 +28,14 @@ jobs:
         run: npm run test:unit
       - name: Build
         run: npm run build:prod
+      - name: Full Page Visual Diff Tests
```
margaree, line 31: Adding this here for now for test purposes. My thought is that we'd actually just want to run this only as a part of the `deploy` action, since it takes a while.

package.json
```diff
@@ -49,7 +49,7 @@
     "markdown-it-anchor": "^8",
     "markdown-it-modify-token": "^1",
     "marked": "^3",
-    "mocha": "^9",
+    "mocha": "^8",
```
margaree, line 52: Mocha 9 causes issues with running visual-diffs locally so locking at 8

pages/assets/demo-snippet.js
```diff
@@ -35,6 +35,10 @@ class ComponentCatalogDemoSnippetWrapper extends LitElement {
 			 * Display demo directly on page rather than in iframe container
 			 */
 			disableIframe: { type: Boolean, attribute: 'disable-iframe' },
+			/**
+			 * Display empty demo iframe
+			 */
+			emptyIframe: { type: Boolean, attribute: 'empty-iframe' },
```
margaree, line 41: Could mark this as for test purposes only since we only would really want it used there?

test/full-page-visual-diff/component-pages.visual-diff.js
```diff
@@ -0,0 +1,41 @@
...
+	componentPages.forEach((component) => {
+		it(component.name, async function() {
+			await page.setViewport({ width: 1500, height: component.height });
+			await page.goto(`${visualDiff.getBaseUrl()}/_site/components/${component.name}`, { waitUntil: ['networkidle0', 'load'] });
+			await page.bringToFront();
+			await page.$eval('div.d2l-component-catalog-content', (elem) => elem.classList.remove('d2l-component-catalog-content-unresolved'));
+			await page.evaluate(() => {
+				const elems = document.querySelectorAll('d2l-component-catalog-demo-snippet');
+				elems.forEach((elem) => {
+					elem.setAttribute('empty-iframe', 'empty-iframe');
```
margaree, line 33: So the loading of the demos was inconsistent due to loading from unpkg, so for now I just made the demos display an empty iframe. Once the perf work is in we could likely change this, so hopefully this is just temporary, since it'd be nice to also capture visual changes to the demos

test/unit/assets.test.js
```diff
@@ -1,5 +1,5 @@
 const assert = require('assert');
```
margaree, line 1: Remainder of file changes is just some reorganization

test/full-page-visual-diff/component-pages.visual-diff.js
margaree: No longer applicable

rollup.config.js
```diff
@@ -21,7 +21,7 @@ const staticFiles = [];
 if (files.includes('@brightspace-ui/htmleditor/htmleditor.js')) {
 	staticFiles.push({
 		src: './node_modules/@brightspace-ui/htmleditor/tinymce/**',
-		dest: '_site/assets',
+		dest: 'assets',
```
margaree, line 24: When the visual-diff tests are run, they're served from the root and so they actually are looking for the assets in `/assets` rather than `/_site/assets`, so this initially puts those assets at the root and then copies them over.

test/full-page-visual-diff/component-pages.visual-diff.js
```diff
@@ -0,0 +1,35 @@
...
+
+describe('Full component pages', () => {
+
+	const visualDiff = new VisualDiff('full-component-page', __dirname);
+
+	let browser, page;
+
+	const componentPages = [
+		{ name: 'alert', height: 4800 },
+		{ name: 'button', height: 5800 }
```
margaree, line 12: Not sure if there are ideas for a better way to do this - currently I have the height hard coded so that we get the right height for each page since they can vary so much.

test/full-page-visual-diff/component-pages.visual-diff.js
```diff
@@ -0,0 +1,35 @@
...
+		await browser.close();
+	});
+
+	componentPages.forEach((component) => {
+		it(component.name, async function() {
+			await page.setViewport({ width: 1500, height: component.height });
+			await page.goto(`${visualDiff.getBaseUrl()}/_site/components/${component.name}`, { waitUntil: ['networkidle0', 'load'] });
+			await page.bringToFront();
+			await page.$eval('div.d2l-component-catalog-content', (elem) => elem.classList.remove('d2l-component-catalog-content-unresolved'));
+			await page.waitForTimeout(1000);
```
margaree, line 30: For some reason the `DOMContentLoaded` in `baseImports.js` isn't being properly triggered to remove the unresolved class. Manually removing it seems to work consistently.

The timeout is because it takes a bit of time for content to properly display after the class above is removed.

test/full-page-visual-diff/component-pages.visual-diff.js
GeorgePlukov: Hmm not sure of a better way to do this. If we set the viewport to some arbitrarily large value (10000?) and then use the `clip` option of `screenshotAndCompare` would that work?

.github/workflows/ci.yml
margaree: Yeah I like that. My biggest concern would be them taking longer than timeout. So far I've run several times here and they've been consistent so currently no worries on test flakiness, but definitely something to keep an eye on.

test/visual-diff/full-pages.visual-diff.js
```diff
@@ -0,0 +1,39 @@
...
+	['status', 'alert', 'html-editor'].forEach((component) => {
+		it(component, async function() {
+			await page.setViewport({ width: 1500, height: 10000 });
+			await page.goto(`${visualDiff.getBaseUrl()}/_site/components/${component}`, { waitUntil: ['networkidle0', 'load'] });
+			await page.bringToFront();
+			await page.$eval('div.d2l-component-catalog-content', (elem) => elem.classList.remove('d2l-component-catalog-content-unresolved'));
+			await page.waitForTimeout(1000);
+
+			// get content height in order to properly size viewport
+			const height = await page.$eval('.d2l-component-catalog-content', (elem) => {
```
margaree, line 28: I had the smoke test visual-diffs that would be in CI just set the height based on the content height, in order to view the side bar in the visual-diffs, whereas I had the deploy tests just do a clip, so they wont capture the sidebar, since that doesn't seem overall needed for all components

margaree (commented):
There was some work done to get the resizing work properly for html editor in the diffs, which overall makes it more stable in the prod site. Should be stable now

GeorgePlukov (approved):

PR #428, opened 2021-11-11
Visual-diff: work on flakiness

margaree:


pages/assets/demo-resizable-preview.js
```diff
@@ -246,6 +247,9 @@ class ComponentCatalogDemoResizablePreview extends FocusVisiblePolyfillMixin(Lit
 			:host([attached][resizable]) playground-preview {
 				border-radius: 10px 0 0 0;
 			}
+			:host([_hideLoadingIndicator]) ::part(preview-loading-indicator) {
```
margaree, line 250: The loading indicator would periodically show up causing visual diff flake. I used local storage in order to be able to hide it in the tests. Open to other ideas here. This seems stable so far

pages/assets/demo-resizable-preview.js
```diff
@@ -472,8 +477,10 @@ class ComponentCatalogDemoResizablePreview extends FocusVisiblePolyfillMixin(Lit
 		}
 	}
 	_handleIFrameMessage(event) {
+		if (!(event.origin.includes('localhost') || event.origin.includes('playground-'))) return;
+
 		const { height, id } = event.data;
-		if (id === this._id && (event.origin.includes('localhost') || event.origin.includes('playground-'))) {
```
margaree, line 476: This got changed when trying a different strategy - can also change back

pages/assets/demo-resizable-preview.js
GeorgePlukov: Maybe we turn off the loading indicator as part of reduced motion, not sure if that is disabled by default in the visual diff tests but we should disable it

margaree: Ah that's a good idea. Yeah I'm not sure if it does - sounds familiar, I'll look into it

margaree: Checked the visual-diff code and looks like it does set `prefers-reduced-motion` (https://github.com/BrightspaceUI/visual-diff/blob/main/helpers/createPage.js#L4) so in theory this should work

GeorgePlukov (approved):

PR #440, opened 2021-11-15
Analytics: don't send google analytics from CI sites or PR visual diff tests

margaree:
It's a bit trickier to prevent the Google Analytics calls from the site deploy visual diffs since we do want to use that built site as the prod site as well as to run the visual-diff tests on.

We could technically do a rebuild after running the visual-diffs on the deploy action, though there would be some risk that something changed between the first build, the visual-diffs, then the second build.

We could also see if there is some way to ignore the visual-diff "user".

Any other ideas?


GeorgePlukov:
I haven't used the GA portal for a while but is it possible within the reporting tool to look at only one subnet on the site? Like just searching for `www.daylight.com`. Honestly though only deploying the GA code when its a prod site seems fine.

GeorgePlukov (approved):

PR #455, opened 2021-11-16
Visual-diff: Add more tests for interactive functionality in demo-tables

margaree:
This PR is mostly to add visual-diffs for testing the edit/interactive states of the attributes table. I noticed a bug (https://github.com/BrightspaceUI/documentation/pull/455#discussion_r750630528) with `interactive` state so there was some work to fix that that including some moving code around.


pages/assets/demo-tables.js
```diff
@@ -118,60 +120,7 @@ export class ComponentCatalogDemoTables extends LitElement {
 	render() {
 		if (!this._componentInfo) return;
 
-		const attributeRows = this._componentInfo.attributes ? this._componentInfo.attributes.map((info) => {
```
margaree, line 121: This was moved out to `_getAttributeRow` for readability within `render` and moved mostly as is (I'll comment on any changes)

pages/assets/demo-tables.js
```diff
@@ -193,29 +142,6 @@ export class ComponentCatalogDemoTables extends LitElement {
 			`;
 		}) : null;
 
-		let demoHeaderButton = html`<div style="width: 2.1rem; height: 1rem;"></div>`; // placeholder for when no button is available
```
margaree, line 196: This was moved out to `_getAttributeHeadings` for readability and mostly moved as is

pages/assets/demo-tables.js
```diff
@@ -225,8 +151,7 @@ export class ComponentCatalogDemoTables extends LitElement {
 								<th>Property</th>
 								<th>Type</th>
 								<th>Description</th>
-								${ !this._editMode ? defaultValueHeader : demoValueHeading }
-								<th @mouseover=${this._handleMouseOver} @mouseout=${this._handleMouseOut}> ${demoHeaderButton}</th>
+								${this.interactive ? this._getAttributeHeadings() : html`<th><div class='d2l-default-value-container'>Default Value</div></th>`}
```
margaree, line 154: So I noticed when `interactive` was false, the edit option would show up on hover but when clicked the column would just disappear. I changed that to remove the edit button option on hover - hopefully that would be what's expected?

pages/assets/demo-tables.js
```diff
@@ -272,6 +197,118 @@ export class ComponentCatalogDemoTables extends LitElement {
...
+		let hasEditedAttributes = false;
+		for (const attribute in this._edited) {
+			if (this._edited[attribute]) {
+				hasEditedAttributes = true;
+				break;
+			}
+		}
+
+		let demoHeaderButton = html`<div style="width: 2.1rem; height: 1rem;"></div>`; // placeholder for when no button is available
+		if (this._editMode) {
```
margaree, line 210: These all function the same, I just switched up the cases to be a bit more intuitive (to me - understood if the previous way is overall more intuitive to the architecture and can definitely change back). I also added `text` to the icon buttons. Otherwise `_getAttributeHeadings` functions as previously.

pages/assets/demo-tables.js
```diff
@@ -272,6 +197,118 @@ export class ComponentCatalogDemoTables extends LitElement {
...
+
+	_getAttributeRow(info) {
+		const infoDefault = info.default ? info.default.replace(/\\"/g, '') : null;
+		let demoType = info.type;
+		let initialValue = infoDefault;
+
+		const defaultShown = infoDefault ? infoDefault.replace(/'/g, '') : html`<div aria-hidden="true">--</div>`;
+		const description = info.description ? unsafeHTML(this._parseMarkdown(info.description)) : '';
+
+		if (!this.interactive) {
```
margaree, line 260: The only change in here is to add an early exist case for when `interactive` is false which removes the edit state option.

test/visual-diff/demo-tables.visual-diff.html
```diff
@@ -4,7 +4,8 @@
 		<script type="module">
 			import '@brightspace-ui/core/components/typography/typography.js';
 			import '../../pages/assets/demo-tables.js';
-			const validInfo = {
```
margaree, line 7: These are mostly renames and removal of `properties` to simplify since we don't use `properties`

test/visual-diff/demo-tables.visual-diff.js
```diff
@@ -23,39 +23,58 @@ describe('d2l-component-catalog-demo-tables', () => {
...
+	[
+		'no-tag',
+		'invalid-tag',
+		'valid-tag',
+		'interactive',
+		'interactive-edit',
+		'no-slots',
+		'hide-slots',
+		'hide-events'
+	].forEach((testName) => {
```
margaree, line 35: The existing visual-diffs were just moved into this `forEach` with the addition of `interactive-edit`

test/visual-diff/demo-tables.visual-diff.js
```diff
@@ -23,39 +23,58 @@ describe('d2l-component-catalog-demo-tables', () => {
...
+		});
 		const rect = await visualDiff.getRect(page, '#interactive');
 		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false, clip: rect });
 	});
 
-	it('No slot items', async function() {
-		const rect = await visualDiff.getRect(page, '#no-slots');
-		await visualDiff.screenshotAndCompare(page, this.test.fullTitle(), { captureBeyondViewport: false, clip: rect });
-	});
+	describe('interactive edit', () => {
```
margaree, line 50: Added tests to capture states of interactive table

pages/assets/demo-tables.js
GeorgePlukov: Definitely a bug, thanks for adding this!

GeorgePlukov (approved):

PR #462, opened 2021-11-18
Minor bug fixes

GeorgePlukov:
1. prevents the `px` from overflowing its container on large screen sizes
2. Allow using escape to blur focus from the resizable slider
3. Prevent the demo from sending height event on subsequent loads (changing attributes was causing it to resize and look kinda bad)


pages/assets/demo-resizable-preview.js
```diff
@@ -306,7 +307,7 @@ class DaylightDemoResizablePreview extends FocusVisiblePolyfillMixin(LitElement)
 					demoEl.classList.remove('hide');
 
 					if (${this._firstLoad}) demoEl.setAttribute('data-first-load', 'data-first-load');
-					${this.autoSize ? `
+					${this.autoSize && this._firstLoad ? `
```
GeorgePlukov, line 310: Not sure if some demos were relying on this, I noticed when clicking attributes quickly there would be a small graphical shift, depending on the demo seems like it could be more. We could technically add this as a parameter if only certain demos need it

margaree: Yeah I think this is likely fine to change. The amount the content could move around could get irritating. The only demo I can think of that would be semi-affected would be text area if rows are added, but you can still scroll to see the whole demo.

pages/about/whats-new.md
```diff
@@ -19,6 +19,9 @@ This page will stay up-to-date with changes to components (e.g., new component p
 ## daylight.d2l.dev
 
 ### 2021-11-15
+- Bug fixes:
```
margaree, line 22: This should be under a different date

margaree (commented):
I was just curious more as to the "why" behind the escape blur change - it seems that the slider still does have focus, since if I tab again it goes to the next tab stop, it's just that the celestine color is removed and the px number eventually disappears. Are there any accessibility concerns around removing focus appearance on a focused element?

pages/about/whats-new.md
GeorgePlukov: Oh I thought this was for the week of the 15th. Are we going to add daily entries?

margaree: Ohh I had thought they were daily. Weekly is likely enough granularity. Maybe we should just indicate that somewhere? Like `Week of 2021-11-15` instead?

GeorgePlukov:
@geurts filed a defect for it [US130956](https://rally1.rallydev.com/#/15545167705d/workviews/collection?Name=&Owner=%2Fuser%2F10152819759&PortfolioItem=%2Fportfolioitem%2Ffeature%2F417115917556&Project=%2Fproject%2F15545167705&collectionView=1bda5c7f-5867-4e6c-bde6-6ca8924b8801&detail=%2Fuserstory%2F607023100131&fdp=true&itemRef=%2Fportfolioitem%2Ffeature%2F417115917556&iteration=u&parentRef=%2Fportfolioitem%2Ffeature%2F417115917556&previousTokens=%2F15545167705d%2Fsearch%3Fdetail%3D%252Fportfolioitem%252Ffeature%252F6a639767-4cd2-4ebe-8096-25f9a3d45083%26keywords%3Dcomponent%2520catalog&project=%2Fproject%2F15545167705&rankTo=BOTTOM&ref=%2Fportfolioitem%2Ffeature%2F417115917556&relationship=userstories) it seemed to make sense to remove the pixel/blue styling but you are right the focus doesn't necessarily jump somewhere meaningful

pages/about/whats-new.md
GeorgePlukov: Yeah makes sense 👍

GeorgePlukov: @margaree Lets stick with the single day format, it'll be annoying for now since we'll have to add a new title pretty much every PR, but if at some point we add versioning for the site we can just collect things under the unreleased section

GeorgePlukov:
@margaree discussed it with Jeff and he agreed to not include the escape functionality. I also took a look at the slider in the primary-secnodary template and it does not support it. Nice call

margaree (approved):
Sounds good! Just need to remove that from the "what's new" page then and then this LGTM.

PR #464, opened 2021-11-19
Reorganize tests into utils and components

margaree:
This splits out the tests into ones for the node script utilities and separately ones for components, then uses the web test runner for the components one in order to be consistent with other component tests.


.gitignore
```diff
@@ -1,11 +1,11 @@
...
-assets/
-node_modules/
-playground-build/
-pages/_includes/manifest.json
-pages/_includes/manifest-wc.json
-pages/assets/manifest-wc.mjs
-pages/components/imported/
-test/**/screenshots/current/
-test/**/screenshots/golden/
+/assets/
```
margaree, line 1: I was having troubles with gitignore when I moved the utils.js file, and found that it was fine with I used `/assets/` instead of `assets/`, so I switched all to use that formatting for consistency/to avoid ambiguity.

GeorgePlukov (approved):
LGTM. only suggestion is maybe removing the `unit-` from the test folder name, I think I would expect them to be unit tests by default and maybe labelled otherwise?

PR #466, opened 2021-11-19
Rename pages/assets to pages/_components and pages/_helpers

margaree:
There will be changes in core to use `autoOpen` in the `docs: demo...` comment rather than actually importing the file.

If this seems okay, I'll need to go through the core dropdown etc components and fix those to use `autoOpen`, then bump core in this PR.


pages/_components/demo-snippet.js
```diff
@@ -16,6 +16,14 @@ const hiddenContentRegex = new RegExp('(<!-- docs: start hidden content -->)(.*?
 class DaylightDemoSnippetWrapper extends LitElement {
 	static get properties() {
 		return {
+			/**
+			 * Import the open-first-item-helper to toggle open the first child demo (e.g., dropdown)
+			 */
+			autoOpen: { type: Boolean, attribute: 'auto-open' },
+			/**
+			 * Default values for demo attributes. Formatted as a stringified object.
```
GeorgePlukov, line 24:
```suggestion
			 * Enable this attribute to allow the demo preview to resize to the size of the first child demo (e.g. d2l-button)
```
I think this comment was wrong before

pages/_data/assets.js
```diff
@@ -1,12 +1,12 @@
 /* global require, module, process */
 module.exports = {
-	getPath: (assetName, nodeModulePath) => {
+	getPath: (assetName, nodeModulePath, isComponent) => {
```
GeorgePlukov, line 3: optional: We could split this into `getNodeModulePath`, `getAssetPath`, and `getComponentPath` if we wanted to avoid the null parameter

GeorgePlukov (approved):

margaree:
Note to self:
- [ ] update core to use `autoOpen` before merging

PR #468, opened 2021-11-22
Unit tests: demo-tables: Initial work to add component tests

margaree:
Also more consistency with using properties term in demo-tables file


pages/assets/demo-tables.js
```diff
@@ -74,17 +74,17 @@ export class DaylightDemoTables extends LitElement {
 			h2.d2l-heading-3 {
 				margin-bottom: 0.5rem;
 			}
-			table.d2l-attribute-table thead tr:first-child th:last-child {
+			table.d2l-properties-table thead tr:first-child th:last-child {
```
margaree, line 77: I switch to use `properties` instead of `attributes` when referencing the table since we call it the properties table but seem to be using attribute and properties interchangeably. I left `attributes` when the component info file was involved.

pages/assets/demo-tables.js
```diff
@@ -198,118 +198,6 @@ export class DaylightDemoTables extends LitElement {
 		this._edited =  { ...this._edited, [details.name]: !details.reset };
 	}
 
-	_getAttributeHeadings() {
```
margaree, line 201: This was just renamed/moved

pages/assets/demo-tables.js
```diff
@@ -376,11 +264,123 @@ export class DaylightDemoTables extends LitElement {
...
+		};
+		const iconItemClasses = {
+			'd2l-value-column-hover': this._mousedOver,
+		};
+		const propertyRowClasses = {
+			'd2l-edited-row': isEdited
+		};
+		const isEditedDot = isEdited ? html`<d2l-icon class="d2l-edited-dot" icon="tier1:dot"></d2l-icon>` : null;
+		return html`
+			<tr class=${classMap(propertyRowClasses)} data-property="${info.name}">
```
margaree, line 323: `data-property` was added to make testing easier.

test/unit-components/demo-tables.test.js
```diff
@@ -0,0 +1,226 @@
...
+						break;
+					}
+					case 'boolean': {
+						emptyDefault = false;
+						break;
+					}
+				}
+
+				[
+					{ editMode: false, default: false, demoDefaults: null, expected: '--' },
```
margaree, line 114: For each attribute type, tests when in edit mode vs default mode what the value in the cell is depending on if the property has a default or not and if there are demo override defaults or not.

margaree:
This did something weird on rebase so I'm just going to open a new PR.

PR #471, opened 2021-11-23
Unit tests: Navigation

margaree:
This moves the page order sorting out to the 11ty build then adds unit tests for that.


tools/eleventy-utils.js
```diff
@@ -25,6 +25,31 @@ module.exports = {
...
 		return result;
 	},
+	handleCollectionChildren: (entry, size) => {
+		const hasChildren = entry.children && entry.children.length > 0;
+
+		if (hasChildren && entry.parent === 'components') {
+			entry.children.sort((a, b) => {
+				if (a.order >= 0 && b.order >= 0) {
+					if (a.order === b.order) {
+						if (a.title < b.title) return -1;
```
GeorgePlukov, line 35: If you want you can do the same `a - b` trick here as we do when we sort attribute names

GeorgePlukov (approved):

tools/eleventy-utils.js
margaree: Does `a - b` work for alphabetical?

PR #485, opened 2021-11-29
Link improvements

margaree:
- Adds ability to tab focus to the hidden heading 2 and 3 links
- Adds links to the headings 3s (were previously left out for consistency since problems with the Properties, Slots etc in the shadowRoot)
- Adds ability to have links for the Properties, Events, and Slots in demo-tables
- Simplifies component hashes (e.g., instead of encoded "Menu [d2l-menu]" would just be "d2l-menu" - it was easier to have consistency using the tag with h2 and properties than use the component official name (e.g., Menu))


pages/_components/demo-snippet.js
```diff
@@ -212,6 +212,12 @@ class DaylightDemoSnippetWrapper extends LitElement {
 		super.update(changedProperties);
 	}
 
+	scrollToHash(id) {
```
margaree, line 215: Used by a new script in `component.njk` for scrolling within shadow root to properties/events/slots

pages/_components/header-link-styles.js
```diff
@@ -0,0 +1,22 @@
+/* stylelint-disable */
+import { css } from 'lit-element';
+
+export const headerLinkStyles = css`
```
margaree, line 4: These were moved from `style.css` so that they could be shared more easily. Biggest change here was to have show/hide done with opacity so that focus would work.

margaree:
The visual diffs in the visual diff PR look correct, just not merging them yet in order to minimize file changes for now.

GeorgePlukov (approved):

PR #491, opened 2021-12-03
Initial work on displaying package version with demos

margaree:
This is mostly to get the infrastructure in place to pass the package version through to the component. It currently displays under the demo or the code (if code is displayed) but likely would need design improvements.

No rush to get this in, feel free to modify if this happens to get prioritized while I'm away, else I'll work more on it when I return.


pages/_includes/layouts/base.njk
```diff
@@ -97,7 +97,7 @@
 		});
 	</script>
 </head>
-<body class="d2l-typography">
+<body class="d2l-typography" data-package="{{ packageName}}@{{ packageVersion }}">
```
margaree, line 100: This is the easiest way to pass the package version through to the demo though I don't love it.

tools/generate-pages.mjs
```diff
@@ -235,8 +239,19 @@ function _getMissingDependenciesList() {
 	let dependencyString = '';
 	return new Promise((resolve) => {
 		repoInstallLocations.forEach((dep) => {
-			exec(`npm list ${dep}`, (error) => {
-				if (error) dependencyString += ` ${dep}`; // dependency is not installed
+			exec(`npm list ${dep} --json --depth 0`, (error, res) => {
+				const resJSON = JSON.parse(res);
+				if (error) {
+					// dependency is not installed
+					if (resJSON && resJSON.dependencies && resJSON.dependencies[dep]) dependencyString += ` ${resJSON.dependencies[dep].from}`; // use daylight package-lock version rather than latest
```
margaree, line 246: I noticed that without this, latest would get installed rather than the version in package-lock

tools/generate-pages.mjs
```diff
@@ -21,6 +21,7 @@ const FILENAME_ROLLUP_COMPONENTS = 'rollup-files-generated-wc.js';
 const FILENAME_ROLLUP_PAGE = 'rollup-files-generated-page.js';
 
 const componentIssues = {},
+	dependencyVersions = {},
```
margaree, line 24: We currently put the package (e.g., `brightspace-ui/core`) in the github issue as `baseInstallLocation`. The changes to this file do the following:
- when installing dependencies, adds them to this `dependencyVersion` object
- when generating the front matter, adds `baseInstallLocation` to that (in the utils file)
- if `baseInstallLocation` is in `dependencyVersions`, adds the `dependencyVersion` to the front matter as `packageVersion`

margaree:
We might want to consider implementing a way to not print the version for tests, as this will cause the visual diff goldens to change every time there's a version bump.

pages/_components/demo-snippet.js
```diff
@@ -193,6 +195,7 @@ class DaylightDemoSnippetWrapper extends LitElement {
 				</div>
 				${ !this.hideCode ? html`<pre class="language-${this.language}"><code class="language-${this.language}">${unsafeHTML(this._highlightedCodeSnippet)}</code></pre>` : null }
 			</div>
+			<div class="d2l-body-compact">Package version: ${this._version}</div>
```
GeorgePlukov, line 198: I know this is a first pass but do we potentially want to overlay it on the demo portion of the component? If it was a light gray it might be prominent enough to read but still sit nicely in the corner

pages/_includes/layouts/base.njk
margaree: Good idea! Should be possible

pages/_components/demo-snippet.js
```diff
@@ -169,6 +173,12 @@ class DaylightDemoSnippetWrapper extends LitElement {
 		super.firstUpdated(changedProperties);
 
 		if (this.disableIframe) this.hideDemo = true;
+
+		const tagInfo = components[this.tagName];
```
margaree, line 177: In order to use `custom-elements-all` to get the version info for the demo and to display it in the demo, it made the most sense to move usage of `custom-elements-all` up to `demo-snippet` and then have component info and version sent to the applicable components.

pages/_components/demo-snippet.js
```diff
@@ -193,8 +204,10 @@ class DaylightDemoSnippetWrapper extends LitElement {
 				</div>
 				${ !this.hideCode ? html`<pre class="language-${this.language}"><code class="language-${this.language}">${unsafeHTML(this._highlightedCodeSnippet)}</code></pre>` : null }
 			</div>
+			${ this._componentVersion && this.disableIframe ? html`<div class="d2l-body-small">Package version ${this._componentVersion}</div>` : null }
```
margaree, line 207: Positioning for `disable-iframe` case is a bit strange so I put it below the demo code for now.

tools/generate-pages.js
```diff
@@ -40,7 +41,7 @@ const octokit = new Octokit({
 });
 
 function _copyCustomElements(repos) {
-	const tags = [];
+	const tags = {};
```
margaree, line 44: I turned the `custom-elements-all` content into an object for faster lookup.

margaree: If this is too disruptive to the search work I can change it back.

GeorgePlukov: Good idea! I can change the search script up a bit to accommodate

margaree:
Visual diffs are being weird on this one. Wondering if it's because the PR is from before and after some version changes. I'm going to open a new PR in draft while I figure this out.

PR #494, opened 2021-12-03
switching to ESM-by-default & upgrading actions

dlockhart:
This accomplishes two things that both needed to happen together:

1) Makes this project use ES Modules by default

Mocha 9 requires it, but this repo is tricky because Eleventy doesn't support ESM (yet). So for non-Eleventy things, I've converted CommonJS stuff (like `require()` and `__dirname`) to the ESM equivalents. For things you had that were already ESM (using a `.mjs` extension), those are now `.js`. And for stuff that had to stay CommonJS for Eleventy they have the `.cjs` extension.

2) New Visual Diff dependencies

By switching to the new `main` branch of the actions repo, you pick up the latest visual-diff, mocha@9 and puppeteer@12 -- this will likely case minor visual differences.


.eslintignore
```diff
@@ -1,2 +1,5 @@
-!.eleventy.js
+!.eleventy.cjs
+!.github
 _site
+assets
```
dlockhart, line 4: When linting after a build I noticed it was linting some of the generated directories. I think this should now be correct.

.github/actions/update-labeled-issues-list/update-issue-list.js
```diff
@@ -1,11 +1,11 @@
...
+import { writeFileSync } from 'fs';
 
 const listPath = process.env['LIST_PATH'];
+const __dirname = path.dirname(process.argv[1]);
 
-const chalk = require('chalk');
-const fs = require('fs');
-const path = require('path');
-const issues = require(listPath);
+const issues = await import(listPath);
```
dlockhart, line 8: A little worried about this dynamic import... we'll see.

.github/workflows/ci.yml
```diff
@@ -10,11 +10,7 @@ jobs:
       - uses: Brightspace/third-party-actions@actions/setup-node
         with:
           node-version: '14'
-      - uses: Brightspace/third-party-actions@actions/cache
-        id: cache
-        with:
-          path: '**/node_modules'
-          key: npm-${{ hashFiles('**/package-lock.json') }}
+          cache: 'npm'
```
dlockhart, line 13: The `setup-node` action actually supports caching out of the box now.

package.json
```diff
@@ -1,18 +1,19 @@
...
     "clean": "rimraf _site && rimraf playground-build && rimraf .generated && rimraf pages/components/imported",
     "lint": "npm run lint:eslint && npm run lint:lit && npm run lint:style",
-    "lint:eslint": "eslint . --ext .js,.mjs",
+    "lint:eslint": "eslint . --ext .js,.cjs",
     "lint:lit": "lit-analyzer pages/**/*.js --strict",
     "lint:style": "stylelint \"**/*.js\"",
-    "start": "cross-env NODE_ENV=development npm run build && eleventy && npm run start:servers",
-    "start:servers": "concurrently \"npm run start:playground\" \"npx eleventy --watch\" \"web-dev-server\"",
+    "start": "cross-env NODE_ENV=development npm run build && eleventy --config=.eleventy.cjs && npm run start:servers",
+    "start:servers": "concurrently \"npm run start:playground\" \"npx eleventy --config=.eleventy.cjs --watch\" \"web-dev-server\"",
```
dlockhart, line 16: Unfortunately Eleventy isn't expecting the config to be a `.cjs` file so we need to tell it about this... in 3 places. I could move it into a separate script if you think that'll be helpful.

README.md
```diff
@@ -22,23 +22,29 @@ npm test
...
 This repo uses the [@brightspace-ui/visual-diff utility](https://github.com/BrightspaceUI/visual-diff/) to compare current snapshots against a set of golden snapshots stored in source control.
 
-The golden snapshots in source control must be updated by Github Actions.  If your PR's code changes result in visual differences, a draft PR with the new goldens will be automatically opened for you against your branch.
+The golden snapshots in source control must be updated by the [visual-diff GitHub Action](https://github.com/BrightspaceUI/actions/tree/main/visual-diff).  If a pull request results in visual differences, a draft pull request with the new goldens will automatically be opened against its branch.
 
-If you'd like to run the tests locally to help troubleshoot or develop new tests, you can use these commands:
+To run the tests locally to help troubleshoot or develop new tests, first install these dependencies:
+
+```shell
+npm install @brightspace-ui/visual-diff@X mocha@Y puppeteer@Z  --no-save
```
GeorgePlukov, line 34: Is it worth including a copy-pastable version of this command too or do you imagine versions changing often enough for that to be a headache?

GeorgePlukov (approved):
LGTM Going to take a quick look at the visual diffs

README.md
svanherk: Yeah our thinking was that the versions change often enough (especially puppeteer), so we've included the copy-pastable command in the action here and linked to it: https://github.com/BrightspaceUI/actions/tree/main/visual-diff#current-dependency-versions

PR #499, opened 2021-12-13
Fixes slight upward movement on page load

GeorgePlukov:
We noticed this when we were doing CR on this pre-pilot but it merged before I could get a fix in.


margaree (approved):

pages/_includes/layouts/base.njk
```diff
@@ -84,11 +87,11 @@
 		let showPilotHeader;
 
 		window.addEventListener('DOMContentLoaded', function() {
-			showPilotHeader = window.localStorage.getItem('hide-pilot-alert');
+			hidePilotAlert = window.localStorage.getItem('hide-pilot-alert');
+			const banner = document.querySelector('.d2l-daylight-warning-banner');
```
margaree, line 91: nit: could move this back into the if

PR #504, opened 2021-12-17
Remove html-editor visual diff tests

GeorgePlukov:


GeorgePlukov:
Seems like the HTML editor has the visual diff tests disabled within its own repo, not sure if its worth fighting over

dlockhart (approved):

PR #517, opened 2021-12-22
Side nav: Improve page navigation/reload performance

margaree:


pages/_includes/layouts/base.njk
```diff
@@ -49,19 +86,105 @@
...
+								{% else %} aria-expanded="false"
+								{% endif %}
+								{% if entry.url == page.url %} class="d2l-link d2l-daylight-link-large d2l-daylight-link-selected"
+								{% else %} class="d2l-link d2l-daylight-link-large"
+								{% endif %}>
+								{{ entry.title }}
+								<d2l-icon icon="tier1:arrow-expand-small" class="d2l-daylight-nav-right-icon"></d2l-icon>
+								<d2l-icon icon="tier1:arrow-collapse-small" class="d2l-daylight-nav-down-icon"></d2l-icon>
+							</button>
+							<ul>
```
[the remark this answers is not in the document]
margaree: Yeah, this is basically everything pulled out of the navigation component, which was part of what made it slow to load. I'll try to figure out how to simplify.

pages/about/adding-component.md
```diff
@@ -5,6 +5,7 @@ eleventyNavigation:
   key: adding
   order: 1
   parent: contributing
+  grandparent: about
```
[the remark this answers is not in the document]
margaree: Using the shortcode removes the need for grandparent🎉

.eleventy.cjs
```diff
@@ -8,6 +8,53 @@ const { getPathLocalSpecified } = require('./pages/_data/assets.cjs');
 const path = require('path');
 const pluginTOC = require('eleventy-plugin-toc');
 
+function _generateList(entry, pageUrl, toc) {
```
margaree, line 11: This is pretty much exactly what was in the navigation component, just without things like `classMap`, and modified to use the expand/collapse style

GeorgePlukov: Minor: we might want to pull this into 11ty utils

GeorgePlukov (approved):

margaree:
The latest commit added in the closing of open section when a new section is expanded, so it would be good to have a review on that and also a bit of a test in case of any weirdness.

Expected behaviour is that the section containing the active page stays expanded unless clicked, otherwise sections close when new sections are expanded.

GeorgePlukov:
LGTM, we might want to add some transitions at some point later on

pages/_includes/layouts/base.njk
```diff
@@ -26,21 +26,129 @@
...
+					if (prevSibling.classList.contains('d2l-daylight-nav-expanded')) openSibling = prevSibling;
+					else prevSibling = prevSibling.previousElementSibling;
+				}
+			}
+
+			if (openSibling) {
+				openSibling.classList.remove('d2l-daylight-nav-expanded');
+				openSibling.querySelector('button').setAttribute('aria-expanded', 'false');
+				openSibling.querySelector('ul').style.maxHeight = null;
+				setTimeout(() => { openSibling.querySelector('ul').style.display = 'none'; }, 210); // delay so animation can run before setting
```
margaree, line 58: The animation works by setting or unsetting max-height based on content, then setting display of the child `ul` to none or block in order to show or hide the content in the tab order.

Because max-height could be changing on a child, it needs to be propagated up to the parent to adjust that max-height.

Overall this seems to work, but is fairly complex, and I'm definitely open to suggestions on simplifying this. Other experimenting with just css wasn't as successful.

margaree (commented):
@GeorgePlukov this is ready for re-review when you have time.

pages/_includes/layouts/base.njk
GeorgePlukov: Honestly we probably should've fought back a bit on this functionality, not sure how much value its adding at the cost of increased complexity.

I Can't necessarily think of a better way to implement this other than similar approaches to what you are doing (adding/removing classes and recursing through siblings/ancestors). And if we want that slide style animation I think setting height/max height seems reasonable. 


 It seems to be working well while I was testing it out

PR #525, opened 2021-12-30
Gplukov/local search

GeorgePlukov:


tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+		// this.field('', { boost: 1 });
+
+		docs.forEach(function(doc) {
+			this.add(doc);
+		}, this);
+	});
+}
+
+function main() {
+	const files = scanHTML(HTML_FOLDER);
```
GeorgePlukov, line 308: This scans for all of the html files that will need to be parsed

tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+	let docs = [];
+	console.log('Building index for these files:');
+	for (let i = 0; i < files.length; i++) {
+
+		docs.push(...readHtml(HTML_FOLDER, files[i], i));
```
GeorgePlukov, line 313: This actually parses the files and returns the records depending on the type of page

tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+	}
+
+	docs = docs.filter((a) => a);
+
+	// Write lunr and preview index to file, they'll be hosted along with the site
+	const idx = buildIndex(docs);
```
GeorgePlukov, line 319: Generates the static index and writes it to file

tools/eleventy-utils.cjs
```diff
@@ -59,5 +58,13 @@ module.exports = {
 		entry.size = size;
 		const newSize = size === 'large' ? 'medium' : 'small';
 		if (hasChildren) entry.children.forEach((child) => module.exports.handleCollectionChildren(child, newSize));
+	},
+	slugify: (s) => {
```
GeorgePlukov, line 62: This just got moved to utils so I could reference it from the search code

tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+	// Parse the rest of the component pages content into a generic record
+	const elements = $('.d2l-daylight-content');
+	const previews = [];
+
+	elements.each((index) => {
+		const element = elements[index];
+		previews.push($(element).text().trim());
+	});
+	const id = `${RECORD_TYPE.COMPONENT_CONTENT}-${url}`;
+	const content = previews.map(c => c.trim()).join(' ');
```
margaree, line 226: Looks like this was trimmed on 223. Does it need to be trimmed again? Is there a case where there are multiple occurrences of `.d2l-daylight-content` on a page, or is this more just in case?

tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+	const sectionRecords = [];
+	titles.each((index) => {
+		const title = titles[index];
+		const content = $(title).nextUntil('h2, h3');
+		const titleText = $(title).text().trim();
+		let contentText = '';
+		content.each((index) => {
+			const el = content[index];
+			let text = $(el).text().trim();
+			text = text.replace(/(\r\n|\n|\r)/gm, ' ');
```
margaree, line 253: Would you mind adding a comment here about what this regex is replacing?

pages/_includes/layouts/component.njk
```diff
@@ -20,13 +20,16 @@ permalink: "components/{{fileName}}/"
   window.addEventListener('hashchange', scrollToHash)
   window.addEventListener('DOMContentLoaded', scrollToHash)
 </script>
+<div class="future-enhancements-container d2l-prevent-indexing">
```
margaree, line 23: Can probably leave out `future-enhancements-container` and `edit-page-container` classes if css isn't being used on them just for easier maintainability

tools/build_index.cjs
```diff
@@ -0,0 +1,338 @@
...
+	}
+}
+
+function buildIndex(docs) {
+	return lunr(function() {
+		this.ref('id');
+		for (let i = 0; i < SEARCH_FIELDS.length; i++) {
+			this.field(SEARCH_FIELDS[i].slice(0, 1));
+		}
+		// this.field('', { boost: 1 });
```
margaree, line 299:
```suggestion
```

tools/search.test.js
```diff
@@ -0,0 +1,21 @@
+
+// import { describe, expect, it, } from '@open-wc/testing';
```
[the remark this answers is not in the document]
margaree: Makes sense to me, though it might also be possible to split out some of the logic in the record generation into smaller functions that don't rely on cheerio that could be unit tested. e.g., the map functions that generates `propertyRecords` or `slotRecords` don't rely on cheerio

tools/generate-pages.js
```diff
@@ -15,6 +18,7 @@ const DIR_IMPORTED_SCREENSHOTS = path.join(DIR_IMPORTED_COMPONENTS, 'screenshots
 const IMAGEPATH = path.join(__dirname, '../node_modules/@brightspace-ui/core/components/icons/images');
 
 const FILENAME_CUSTOM_ELEM = 'custom-elements-all.js';
+const FILENAME_CUSTOM_ELEM_COMMON = 'custom-elements-all.cjs';
```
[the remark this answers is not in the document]
GeorgePlukov: I can give that a shot, I hadn't tried importing the cjs version there

tools/generate-pages.js
```diff
@@ -1,12 +1,15 @@
...
 import { ISSUE_LABELS, TIERS } from './states.js';
+import dotenv from 'dotenv';
 import { exec } from 'child_process';
 import matter from 'gray-matter';
 import { Octokit } from '@octokit/rest';
 import { parseBody } from './generate-pages-utils.js';
 import path from 'path';
 import { default as publishedComponents } from './component-page-issues.js';
 
+dotenv.config();
```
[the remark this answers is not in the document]
GeorgePlukov: Alright, I'll split this off in to its own PR. I think its worth adding

tools/generate-pages.js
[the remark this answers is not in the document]
GeorgePlukov: I don't think we can mix the file import type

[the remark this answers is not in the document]
margaree: Ok, thanks for looking into it

pages/_includes/layouts/base.njk
```diff
@@ -27,6 +27,7 @@
 	{%- endif %}
 	<script src="{{assets.getHelperPath("base-imports.js")}}" type="module"></script>
 	<script src="{{assets.getComponentPath("navigation.js")}}" type="module"></script>
+	<script src="https://unpkg.com/lunr/lunr.js"></script>
```
margaree, line 30: Would we want to/is it possible to add this to rollup and use that copy rather than rely on unpkg?

GeorgePlukov: I think that would make sense! I'll try and add it this way

GeorgePlukov:
@margaree The code looks quite different from the last time you saw, general notes:
- Moved from cjs back to js (duplicating the slugify function in the process)
- moved functions into utils (+ tests)
- reformatted record generation to remove side effects (now records and previews are returned)
- Moved lunr js into manifest

I think it's a lot cleaner now, let me know if there are any questions

tools/generate-pages.js
```diff
@@ -333,14 +336,23 @@ async function _requestIssues() {
 	return await octokit.paginate(octokit.rest.issues.listForRepo, props);
 }
 
-function _writeJSONToGeneratedFile(data, fileName) {
+function _writeJSONToGeneratedFile(data, fileName, type) {
```
GeorgePlukov, line 339: I should be able to remove these duplicates now too

.eleventy.cjs
```diff
@@ -184,7 +178,7 @@ module.exports = function(eleventyConfig) {
 			}
 		} else if (content.includes('<!-- docs: start hidden content -->')) {
 			inHiddenBlock = true;
-			return '<div style="display: none;">';
+			return '<div class="d2l-hidden-content" style="display: none;">';
```
margaree, line 181: Looks like this might be the only place this class is used, in which case could we just use the `d2l-prevent-indexing` class here as well?

pages/_helpers/base-imports.js
```diff
@@ -7,6 +7,45 @@ import 'regenerator-runtime/runtime';
...
+	const previews = results.map((res) => {
+
+		const record = PREVIEW_LOOKUP[res.ref];
+		if (!record) {
+			console.error(`Record not found: ${res}`);
+		}
+		return { score: res.score, ...record };
+	});
+
+	const slotsAndPropertyResults = previews.filter(({ type }) => (type === RECORD_TYPE.COMPONENT_PROPERTY || type === RECORD_TYPE.COMPONENT_AXE_PROPERTY || type === RECORD_TYPE.COMPONENT_SLOT));
```
margaree, line 43:
```suggestion
	const slotsAndPropertyResults = previews.filter(({ type }) => (type === RECORD_TYPE.COMPONENT_PROPERTY || type === RECORD_TYPE. COMPONENT_A11Y_PROPERTY || type === RECORD_TYPE.COMPONENT_SLOT));
```

tools/build-index.js
```diff
@@ -0,0 +1,273 @@
...
+	cells.each((index) => {
+		const cell = cells[index];
+		previewText = previewText.concat(` ${$(cell).text()} |`);
+		content = previewText.concat($(cell).text());
+	});
+	return [content, previewText];
+}
+
+// super hacky but I think we only label tables with attribute in the first header column if its A11Y based properties
+function isA11YPropertiesTable(headerText) {
```
margaree, line 31: Since this is similar to the other functionality moved to `search-utils`, should this be moved as well?

tools/build-index.js
```diff
@@ -0,0 +1,273 @@
...
+			const $ = cheerio.load(demo);
+			const tagName = $(demo).attr('tag-name');
+
+			if (!tagName) {
+				return;
+			}
+
+			// all demos with properties will hit here, grab the properties from the custom-elements file
+			const codeSnippet = $(demo).text();
+			// todo: this is changed
```
margaree, line 152: Is this comment still needed?

tools/build-index.js
```diff
@@ -0,0 +1,273 @@
...
+		}, this);
+	});
+}
+
+function main() {
+	const filenames = scanHTML(HTML_FOLDER);
+	const docs = [];
+
+	for (let i = 0; i < filenames.length; i++) {
+
```
margaree, line 248:
```suggestion
```

margaree (commented):
Definitely looking cleaner!

I can't comment directly on the file, but looks like `.DS_Store` got committed and I'm wondering if that was needed/if that should be in .gitignore

tools/build-index.js
GeorgePlukov: I could move this one too, I left it here since I thought the comment is more useful in the search index building context

rollup.config.js
```diff
@@ -13,6 +13,7 @@ const componentFiles = [
 	'./pages/_components/icon-library.js',
 	'./pages/_components/status-table.js',
 	'@brightspace-ui/core/components/typography/typography.js',
+	'lunr'
```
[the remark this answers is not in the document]
GeorgePlukov: I can't seem to get lunr imported in the `base-imports` file without it throwing some sort of error. With the  current solution I tried to get  a fair mid ground of including it with our bundle but using lunrs preferred import method still. Definitely would prefer using import syntax

pages/_helpers/base-imports.js
GeorgePlukov: Solid argument for typescript right here

tools/generate-pages.js
```diff
@@ -339,8 +340,9 @@ function _writeJSONToGeneratedFile(data, fileName) {
 	const json = JSON.stringify(data, null, '\t');
 	const fileContent = `/* eslint quotes: 0 */
 
-export default ${json};
```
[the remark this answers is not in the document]
GeorgePlukov: I think my linter keeps auto indenting it for me 🤕

package.json
```diff
@@ -4,6 +4,7 @@
   "type": "module",
   "scripts": {
     "build": "npm run clean && node ./tools/generate-pages.js && npm run build:rollup:wc && npm run build:rollup",
+    "build:index": "node ./tools/build-index.js",
```
margaree, line 7: Can this step be added to `start` so that in dev we can still just do `npm run start` and not have to worry about forgetting to build the index?

GeorgePlukov: Whats the ideal situation here? Should the search index be constantly rebuilding along with the site in the background (not exactly sure the best way to implement this)?  

Doing dev work locally so far I ve been running the build command when needed and the site will automatically reload

margaree: I don't think it would need to be constantly rebuilding, just the first time and maybe mention how to update it in the readme in the `Running` section

GeorgePlukov: I just wanted to clarify since at one point we thought we'd be using an api and definitely wouldnt want to be thrashing that. Now that its a local index I think it could make sense to build and keep up to date, although it would be duplicating a lot of work without adding a cache layer

PR #536, opened 2022-01-07
Development documentation: initial work on structure

margaree:
I added some details for each page on what would be moved. I don't plan on merging this as is - I can branch off this branch to add each page, or just add each one by one separately, leaving this PR in draft.

Let me know any thoughts on this structure. It basically divides development into 2 sections: Best Practices & Web Components.

Other pages that can be added in the future under Best Practices once we have more content for them:
- Performance
- Documentation
- Localization

It's potentially easiest to just use the PR site to preview this for the details on the structure ideas.


pages/developing/best-practices/index.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: layouts/base
+eleventyNavigation:
+  title: Best Practices
```
margaree, line 4: I'm not sure if `Best Practices` is the best name for this section - perhaps `General Concepts` or `Foundations`?

pages/developing/web-components/lit-best-practices.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: layouts/base
+eleventyNavigation:
+  title: LitElement Best Practices & Gotchas
```
margaree, line 4: If this is a "Best Practices" perhaps it would be better under the best practices section (if we keep that name), though it is also web component specific, whereas the other sections are less so.

pages/developing/best-practices/accessibility.md
```diff
@@ -0,0 +1,13 @@
...
+  title: Accessibility
+  key: accessibility
+  parent: best-practices
+---
+
+# Accessibility
+
+https://desire2learn.atlassian.net/wiki/spaces/Accessibility/overview
+
+This page can include an introduction to the importance of accessibility to development at D2L, along with useful links to the Accessibility Hub in Confluence.
```
dlockhart, line 13: Cross-linking to Confluence will be a hurdle if we want to make this site public in the future. I really like our inline accessibility tips within each component's docs... is this adding much beyond that?

pages/developing/best-practices/index.md
dlockhart: What if we didn't have "best practices" at all, but split up the existing content a bit to separate out the "getting started" stuff that you might read once and then move on from the other stuff that you'd refer back to over and over again?

Something like:

- Development
  - Getting Started
    - What are web components? (and why are components good for D2L)
    - Lit
    - Types of web components
    - Component Tiers
  - Creating a Component
    - Before You Build
    - Creating a repo
    - NPM initializer
  - Testing (split the one doc up into several)
    - Linting
    - Unit Tests
    - Accessibility Tests with aXe
    - Visual Diff Testing
    - Cross-browser Testing
  - Advanced
    - Lit Best Practices
    - Performance
    - etc.

GeorgePlukov: Yeah I think it would be nice to have this as `Foundations` Especially since there is another best practices page under Web components

margaree: Great idea! That split flows a lot better

pages/developing/best-practices/accessibility.md
margaree: That's a good point. We can leave it out for now and then re-add if we do have some ideas for differentiating content.

GeorgePlukov (approved):

GeorgePlukov:
The content seems to flow pretty nicely this way, I would say that maybe some of the testing pages look like they won't have enough content to justify a separate page (linting)

margaree:
This has been captured in subsequent PRs so closing this one.

PR #541, opened 2022-01-07
Auto merge package-lock.json update PRs that pass CI

svanherk:
If you want it!  If CI fails it won't merge, and if the site breaks I _think_ you'd find out quickly with the visual-diff full site smoke tests that run after merge.


margaree (approved):
Nice, thanks!

PR #553, opened 2022-01-12
Developing: Creating a Component section, plus some of lit best practices

margaree:


pages/developing/creating-component/before-building.md
```diff
@@ -0,0 +1,144 @@
...
+**Learn More**: [Google Web Fundamentals: Web Components - Composition and Slots](https://developers.google.com/web/fundamentals/web-components/shadowdom#composition_slot)
+
+Recommendations:
+
+- Similar to attributes, strike a balance between flexibility and complexity when choosing which slots to expose
+- To limit what a consumer can put inside a slot, an attribute may be more appropriate. For example, `<my-button text="click me"></my-button>` may be preferable to `<my-button>click me</my-button>` if you want to ensure that nothing other than text can go inside the button.
+
+### Attributes
+
+See [Attributes: Component Architecture](/developing/advanced/lit-best-practices/attributes-properties/#component-architecture).
```
margaree, line 132: Looking to get some feedback on this strategy:
- I divided the lit best practices page by category (have only done the 3 mentioned here so far and only completed the `Attributes & Properties` page just to get more feedback before doing more) since the page would be huge
- I moved the architecture information about that category to the applicable page so that all the information about it was together

GeorgePlukov: It feels a bit weird splitting up the best practices and not have one location where they would still all be viewable from. Im not sure how willing people will be to use the global search for specific content like that and I feel like it may get buried

margaree: Fair. It is a giant document though and I fear that leaving it as a single page would also just be overwhelming. What about instead having something like a `Development Best Practices` section (at same level as `Getting Started`, `Creating a Component`, etc.) and then the pages are within that, so then the pages aren't several levels deep

GeorgePlukov: Yeah avoiding the fourth level of nesting would be amazing. I think it should probably just be called `best practices` since its under developing already. Thinking about it a bit more the "best practices" section probably shouldnt be under advance (unless we want to scare people away)

margaree: @dlockhart wondering if you had any thoughts on this, since it's a change from what was discussed in the spike PR. The TLDR was I thought it would make sense to split out the Lit Best Practices document into pages based on its sections, which could also combine in the component architecture information pertinent to them. If previewing the PR site, the [Attributes & Properties page](https://pr-553-daylight.d2l.dev/developing/best-practices/attributes-properties/) would be the most complete example.

New sections would become:
- Developing
  - Getting Started
    - Web Components 101
    - Lit
    - Types of web components
    - Component Tiers
  - Creating a Component
    - Before You Build
    - Creating a repo
    - NPM initializer
  - Best Practices (could also call this Lit Best Practices but it could be better to leave it generic)
    - Attributes & Properties
    - Events
    - Methods
    - Styles
    - Rendering
    - Lit Gotchas
  - Testing
    - Linting
    - Unit Tests
    - Accessibility Testing
    - Visual Diff Testing
    - Cross-browser Testing
  - Advanced (performance, etc.)

dlockhart: I like that (splitting it up), as I agree that document is really long. Also dropping the "Lit" makes sense, since everything Daylight is currently Lit.

GeorgePlukov (approved):

PR #554, opened 2022-01-13
Fix update-labeled-issues-list action issue with parsing issues file

margaree:


.github/actions/update-labeled-issues-list/update-issue-list.js
```diff
@@ -1,21 +1,20 @@
 import chalk from 'chalk';
+import { default as issues } from '../../../tools/component-page-issues.js';
```
margaree, line 2: This didn't seem to work as a variable and also it seems fine to me to just have it hard coded.

PR #556, opened 2022-01-13
Ignore node_modules in .github

margaree:
See https://github.com/BrightspaceUI/documentation/pull/555/files
Seems like dependencies are being installed in .github and aren't being ignored on commit


GeorgePlukov (approved):
Nice find!

PR #565, opened 2022-01-14
Developing: testing

margaree:


pages/developing/testing/accessibility.md
```diff
@@ -0,0 +1,20 @@
...
+eleventyNavigation:
+  title: Accessibility Testing
+  key: accessibility-testing
+  parent: testing
+  order: 4
+---
+
+# Accessibility Testing
+
+<blockquote>Accessibility testing with aXe and continuous integration with GitHub Actions are setup by default when using the <d2l-link href="/developing/creating-component/npm-initializer/">@brightspace-ui npm initializer</d2l-link>.</blockquote>
```
margaree, line 12: I wanted to add a consistent way across documents to highlight that this was setup by the initializer, so I tried out the blockquotes for this. Thoughts on this? I'd also probably want to tweak the wording to be more consistent between documents.

GeorgePlukov: Is blockquote semantically the right element to use here? This is more of an additional piece of info rather than a direct quote from something that you could cite. I haven't really looked at how other websites do this type of component in the past

margaree: That's a good point - maybe some sort of stylized div would be best. I'll see what patterns we have that could fit this.

GeorgePlukov (approved):

PR #576, opened 2022-01-19
Add GitHub repo link to the bottom of page, put borders between h2 sections

margaree:


.eleventy.cjs
```diff
@@ -40,6 +40,32 @@ module.exports = function(eleventyConfig) {
...
 		return `<div class="d2l-edit-component-page">Suggest an <d2l-link href="${`${repo}edit/${branchName}/${componentPath}`}">edit</d2l-link> for this page</div>`;
 	});
+	eleventyConfig.addNunjucksShortcode('gitHubLink', (repo, devMarkdown, codeDirectory) => {
+		if (!repo) return '';
+
+		const repoNameMatch = repo.match(/github.com\/(.*\/.*)\//);
+		let repoName = repo;
+		if (repoNameMatch.length === 2) repoName = repoNameMatch[1];
+
+		let markdownDirectory = devMarkdown; // e.g., components/buttons/README>md
```
margaree, line 50: The goal is to link to where the code is. Generally (for core) the code is at the path with README.md removed. `forms` is an exception to this, which I'm adding `codeDirectory` to the issue template for. For cases like HTML editor where the code is just at the repo, it will just end up as that.

.eleventy.cjs
```diff
@@ -40,6 +40,32 @@ module.exports = function(eleventyConfig) {
...
+		if (codeDirectory) {
+			markdownDirectory = `tree/main/${codeDirectory}`;
+		} else if (devMarkdown.includes('/')) {
+			const markdownMatch = devMarkdown.match(/(.*)\/.*.md/); // match would be components/buttons
+			if (markdownMatch && markdownMatch.length === 2) markdownDirectory = `tree/main/${markdownMatch[1]}`;
```
margaree, line 55: I think there was another case where we made the assumption that `main` was the primary branch but I can't remember where/if there was anything added to handle the case where it might be something else.

tools/eleventy-utils.cjs
```diff
@@ -48,6 +48,36 @@ module.exports = {
...
+		let result = demoSnippet.split('-->')[0];
+		if (!result.includes('<!-- docs: demo')) return undefined;
+
+		result = result.replace(/<!-- docs: demo( live)?( code)?/, '');
+		result = result.trim();
+		result = result.replace(/(\S+): ?(\S+)/g, '"$1":"$2"');
+		result = result.replace(/\n|\r|\r\n| /g, ',');
+		return JSON.parse(`{${result}}`);
+	},
+	getGitHubCodeInfo: (repo, devMarkdown, codeDirectory) => {
```
GeorgePlukov, line 61: Probably would be ideal to split this into a function for repoName and one for repoUrl. It would definitely make the function name more obvious to what they are doing, and make the tests a bit cleaner.

GeorgePlukov:
> I think there was another case where we made the assumption that main was the primary branch but I can't remember where/if there was anything added to handle the case where it might be something else.

A default branch parameter can be passed in to the editPage shortcode. I don't remember the exact pages it was necessary for but might be worth adding here too

https://github.com/BrightspaceUI/documentation/pull/576/files#diff-52808709ca65dcdceff6547d303db5b2a4b8425d78460a1f388491c836798f78R36

GeorgePlukov (approved):

PR #581, opened 2022-01-24
Status Tables: Update with new states

margaree:
Based on state changes:

| Development State | Stability State | Display |
|---|---|---|
| Completed | N/A | Stable |
| In Progress | Stable | Stable, in development |
| In Progress | Not Stable | Not Stable, in development |
| Inactive | Not Stable | Not Stable, inactive |
| Not Planned | | Not Planned |
| On the Backlog | | On the Backlog |
| Deprecated | | Deprecated |


margaree:
Visual diffs look as expected, will wait to merge until after review.

pages/_components/status-table.js
```diff
@@ -2,24 +2,35 @@ import '@brightspace-ui/core/components/colors/colors.js';
...
 					${CC_STATES.STABLE}
 				</div>`;
 		case DEV_STATES.IN_PROGRESS:
-			return CC_STATES.IN_DEVELOPMENT;
+			if (stabilityStatus === STABILITY_STATES.STABLE) {
+				return html`
+					<div class="d2l-daylight-status-stability">
+						<div class="d2l-daylight-status-stable">
+							<d2l-icon icon="tier1:check" style="color: var(--d2l-color-olivine);"></d2l-icon>
+							${CC_STATES.STABLE}</div>, ${CC_STATES.IN_DEVELOPMENT}
```
GeorgePlukov, line 27: Does the HTML have to be formatted like this? Not easy to see the closing div here.

It also may be worth keeping the same class name/div structure between the different return cases here if its not necessary. Seems like one returns a div within a div and the other is one flat div with both classnames?

pages/_components/status-table.js
```diff
@@ -92,7 +109,9 @@ class DaylightStatusTable extends LitElement {
 				? html`<d2l-link href="${linkHref}">${component.name}</d2l-link>`
 				: html `<d2l-link class="d2l-daylight-status-table-name" href="${linkHref}">${component.name}<img src="/img/${this.tier}-icon.svg" /></d2l-link>`;
 
-			const status = getStatus(component.development);
+			// const status = getStability(component.stability || component.development);
```
GeorgePlukov, line 112: Can probably remove these

pages/_components/status-table.js
```diff
@@ -2,24 +2,35 @@ import '@brightspace-ui/core/components/colors/colors.js';
...
-			return CC_STATES.IN_DEVELOPMENT;
+			if (stabilityStatus === STABILITY_STATES.STABLE) {
+				return html`
+					<div class="d2l-daylight-status-stability">
+						<div class="d2l-daylight-status-stable">
+							<d2l-icon icon="tier1:check" style="color: var(--d2l-color-olivine);"></d2l-icon>
+							${CC_STATES.STABLE}</div>, ${CC_STATES.IN_DEVELOPMENT}
+					</div>`;
+			} else {
+				return html`${STABILITY_STATES.NOT_STABLE}</div>, ${CC_STATES.IN_DEVELOPMENT}`;
```
GeorgePlukov, line 30: Should not stable have an `X` icon? Feels weird to only use icons for the positive case

GeorgePlukov (commented):
I'm not personally a huge fan of the difference

pages/_components/status-table.js
margaree: Yeah I'm not totally sold on the appearance of all the states. I initially did have `not stable` be orange with a caution icon, but Jeff thought that called it out too much. Definitely can iterate on this.

GeorgePlukov (approved):

PR #596, opened 2022-01-28
Demos: Advance properties infrastructure

margaree:
Initial infrastructure to filter advanced properties to be hidden.


pages/_components/demo-tables.js
```diff
@@ -280,21 +357,25 @@ export class DaylightDemoTables extends ScrollShadowRootMixin(LitElement) {
...
 
 		if (!this.interactive) {
 			return html`
-			<tr>
-				<th scope="row"><span class="d2l-property-name">${info.name}</span></th>
-				<td class="d2l-design-system-component-type">${demoType}</td>
-				<td class="d2l-property-description">${description}</td>
-				<td class="d2l-default-value-row">${defaultShown}</td>
+			<tr class="${classMap(classes)}" aria-hidden="${isAdvanced && !this._showAdvancedProperties}">
+				<th scope="row"><div><span class="d2l-property-name">${info.name}</span></div></th>
```
margaree, line 375: Content needed to be wrapped in divs in order to be able to use `max-height` and `overflow` on the data

pages/_components/demo-tables.js
```diff
@@ -100,30 +109,73 @@ export class DaylightDemoTables extends ScrollShadowRootMixin(LitElement) {
...
+		if (this._componentInfo && this._componentInfo.attributes) {
+			this._componentInfo.attributes.forEach((attribute) => {
+				if (attribute.description.startsWith(ADVANCED)) {
+					this._hasAdvancedProperties = true;
+					attribute.description = attribute.description.split(ADVANCED)[1];
+					this._advancedProperties.push(attribute);
+				} else {
+					this._initialProperties.push(attribute);
+				}
+			});
```
dlockhart, line 166: If you're not expecting the values in `_advancedProperties` and `_initialProperties` to change after this code runs, I don't think they actually need to be Lit properties -- instead they could just be private class variables. The only reason to make them properties would be if you'd like an automatic re-render when their value changes, but since they're arrays pushing/popping doesn't trigger a re-render anyway.

Similarly, `_hasAdvancedProperties` isn't really serving a purpose as a property beyond being a shortcut for `this._advancedProperties.length > 0` and how changing it here to `true` is what's likely triggering a re-render. Could replace it with something like: `if (this._advancedProperties.length > 0) { this.requestUpdate(); }`.

dlockhart (approved):

PR #611, opened 2022-02-02
Search: Front end changes for linking with back end

margaree:
I have all the changes working in the branch `search-backend-frontend` but it would be a huge PR so I'm trying to split out some pieces where possible.

This PR is for the remaining front end changes to link with the back end. Currently it does still just use sample data here, but once the back end changes go in, this will work.

Working search can be seen at https://pr-613-daylight.d2l.dev/


pages/_components/search-results.js
```diff
@@ -3,6 +3,36 @@ import '@brightspace-ui/core/components/list/list-item.js';
...
+		primaryContent = RECORD_TYPE_CAT_NAME[item.type];
+		secondaryContent = headersDifferent ? `${item.parentHeader} > ${item.sectionHeader}` : item.parentHeader;
+		supportingContent = html`<code>${item.preview.name || 'Default'}</code>: ${item.preview.description}`;
+	} else if (item.sectionHeader && headersDifferent) {
+		// only display page header as section header if it's different from section header
+		primaryContent = item.sectionHeader;
+		secondaryContent = item.parentHeader;
+	}
+
+	if (supportingContent && supportingContent.length >= MAX_PREVIEW_CHARS) supportingContent = `${supportingContent}...`;
```
margaree, line 25: We crop the preview content when generating the preview within the search backend code, so this adds the ellipses here. We could also crop it here, but I'd be worried there could be massive previews.

pages/_components/search.js
```diff
@@ -84,9 +109,12 @@ export class DaylightSearch extends LitElement {
 	}
 
 	_handleSearch(e) {
+		this._opened = false;
 		this._searchTerm = e.detail.value;
-		this._opened = true;
 		this._search(this._searchTerm);
+		setTimeout(() => {
+			this._opened = true;
```
margaree, line 116: This is so that when a search occurs, dropdown sizes correctly to have scrolling if needed. I couldn't get this working without the `setTimeout` (e.g., tried `requestAnimationFrame` and some combinations of `await this.updateComplete`) - not sure if there's another way I've forgotten/don't know about

pages/_components/search.js
```diff
@@ -95,42 +123,42 @@ export class DaylightSearch extends LitElement {
...
 		}, {
 			id: 'item2',
 			u: '/developing/getting-started/component-tiers/#why-tiers',
-			h: 'Why Tiers',
-			cs: 'developing',
-			type: 'cc',
-			pU: '/developing/getting-started/component-tiers/',
-			c: 'Tiers allow us to have clear, well-defined boundaries between components that adhere to a certain level of quality and robustness and those that are more experimental in nature.'
+			h: 'Component Tiers',
+			sh: 'Why Tiers',
```
margaree, line 146: These have been updated with what will be the object keys with the backend changes.

pages/_helpers/base-imports.js
```diff
@@ -8,45 +8,6 @@ import 'regenerator-runtime/runtime';
 import { customTableStyles } from '../_components/table-style.js';
 import { headerLinkStyles } from '../_components/header-link-styles.js';
 import { linkStyles } from '@brightspace-ui/core/components/link/link.js';
-import { RECORD_TYPE } from '../../tools/search-utils.js';
```
margaree, line 11: All this won't be needed anymore once the search frontend is in. This would technically break a workflow if someone is using this currently, but I doubt it.

pages/_includes/css/style.css
```diff
@@ -58,6 +58,9 @@ d2l-alert.d2l-daylight-warning-banner {
 	margin-top: 2.7rem;
 	padding-top: 1.8rem;
 }
+.d2l-daylight-content h2.d2l-heading-2:focus-visible {
+	outline: none;
```
margaree, line 62: The focus visible outline appears when a page is navigated to using the search link - not sure if that's what we want. It doesn't look great currently, so I removed it for now

tools/build-index.js
```diff
@@ -191,7 +190,7 @@ const parseContentPage = (url, $) => {
 			text = text.replace(/(\r\n|\n|\r)/gm, ' ');
 			contentText += ` ${text}`;
 		});
-		const id = `${RECORD_TYPE.MARKDOWN_CONTENT}-${url}#${slugify(titleText)}`;
+		const id = `${RECORD_TYPE.CONTENT}-${url}#${slugify(titleText)}`;
```
margaree, line 193: I've made some changes such that we don't need to differentiate between markdown_content and component_content, so it's now just content.

pages/_components/search.js
dbatiste: If I understand correctly, the search is synchronous because it is searching a pre-generated index, and is loaded during `firstUpdated`.  Is it safe to assume that that index will be loaded by the time the user can interact?

Even if the actual search is safely synchronous, the `_search(term)` method assigns to `_results` at the end, however Lit property updates are batched and effectively these updates and the rendering become async - and rAF may not be sufficient.  Another way to tackle this may be to wait for `this.updateComplete;` - either after calling `this._search(this._searchTerm);` or by returning it from `_search(term)` and changing to `await this._search(this._searchTerm);`. Just a thought.

dbatiste: Oh, I see you already tried `await this.updateComplete`.  Hmmm.  That's my thoughts on what's happening here.  Maybe by waiting for `updateComplete` just means Lit has processed the render, but the browser hasn't necessarily applied the changes yet. 🤔

dbatiste (approved):

pages/_components/search.js
margaree: Yup it should be fairly safe to assume the index would be loaded before the user interacts.

Yeah it's strange - I've tried `updateComplete`s all around the `_search` and cannot get it working. I think I'll leave it as is for now and revisit.

PR #659, opened 2022-02-09
Search: "s" key shortcut

margaree:
Uses local storage to track if use wants s key to trigger focus on search and adds a switch to toggle this preference. Shows a tooltip on first focus with information on the "s" key.


pages/_components/search.js
```diff
@@ -60,6 +69,10 @@ export class DaylightSearch extends LitElement {
 	async firstUpdated(changedProperties) {
 		super.firstUpdated(changedProperties);
 
+		this._searchInput = this.shadowRoot.querySelector('d2l-input-search');
+		const ignoreShortcut = window.localStorage.getItem('ignore-search-shortcut');
```
annabelsegalld2l, line 73: Do we want this to be d2l-daylight-ignore-search-shortcut in case some other local storage is already using that name?

pages/_components/search.js
```diff
@@ -81,31 +94,50 @@ export class DaylightSearch extends LitElement {
...
+							text="Enable 's' shortcut key"
+							text-position="start">
+						</d2l-switch>
+					</div>
 				</d2l-dropdown-content>
 			</d2l-dropdown>
 		`;
 	}
 
+	async focus() {
```
annabelsegalld2l, line 135: Does this need to be async? None of the lines have `await`

pages/_components/search.js
```diff
@@ -121,6 +153,18 @@ export class DaylightSearch extends LitElement {
 		}, 0);
 	}
 
+	_handleSwitchToggle(e) {
+		window.localStorage.setItem('ignore-search-shortcut', !e.target.on);
```
annabelsegalld2l, line 157: This is a little complicated to follow so I just want to double-check: 

`e.target.on` -> `ignore-search-shortcut = false` -> switch will be on

pages/_components/search.js
margaree: Ah yup! Forgot to remove that

pages/_components/search.js
margaree: Yup that's correct.

Yeah it's a little complex just because of the initial state of the local storage variable being undefined, though we could have an undefined state equal to use shortcut instead, it's just a bit reverse of the usual undefined being similar to false. I can play around with this a bit.

PR #667, opened 2022-02-10
Search: Use Menu instead of List, arrow keys, sticky header

margaree:
This PR convert list to use a menu and get the colors and spacing to design spec.


pages/_components/search-result.js
```diff
@@ -0,0 +1,195 @@
...
+					--d2l-menu-border-color-hover: var(--d2l-color-amethyst-plus-1);
+					padding: 0.9rem 0.9rem;
+				}
+				:host {
+					display: block;
+				}
+				:host([hidden]) {
+					display: none;
+				}
+				:host(:hover:not(:focus)) {
```
margaree, line 70: We want the border colors to change only on focus, not hover

margaree (commented):
Visual diffs look as expected

PR #697, opened 2022-02-14
Search: Move location in side nav, improve scroll behaviour, open on focus and add empty state

margaree:
Additional changes based on feedback from Jeff. This branch is on top of the `search-list-arrow-keys` branch.


pages/_includes/layouts/base.njk
```diff
@@ -72,6 +72,19 @@
 			_expandParents(parentUl, additionalHeight);
 		}
 
+		function handleNavScrolling() {
+			const nav = document.querySelector('nav');
+			if (!nav) return;
+
+			const searchContainer = document.querySelector('.d2l-daylight-search-container');
```
annabelsegalld2l, line 79: Should this const be inside the listener? In case it gets added to the DOM later.

pages/_components/search.js
```diff
@@ -148,42 +155,39 @@ export class DaylightSearch extends LitElement {
...
+		this._hasResults = false;
+		this._results = undefined;
 		this._searchTerm = e.detail.value;
 		if (!this._searchTerm) return;
 
 		this._search(this._searchTerm);
-		setTimeout(() => {
-			this._opened = true;
-		}, 0);
+		this._dropdownContent.resize();
```
annabelsegalld2l, line 170: A bit confused about this line. Why is the dropdown content getting resized if this._opened is false?

margaree: Generally the dropdown will be opened when the user searches, since it opens on focus, so we want to resize based on the search content. There is the potential case where the user has used escape to close the dropdown, then performs a search, then either hits enter or clicks the search icon to search, where `this._opened` would initially be false. I can likely just move this to after this._opened is true.

annabelsegalld2l (approved):

PR #713, opened 2022-02-23
Branding Updates: Color changes

margaree:
This adds the marketing palette colors for the grays, oranges, and greens, then uses those for several color updates. Also updates the logo.


pages/_helpers/colors-marketing.js
```diff
@@ -0,0 +1,43 @@
...
+			--d2l-color-orange-40: #ff5f13;
+			--d2l-color-orange-50: #ea4711;
+			--d2l-color-orange-60: #ce3517;
+			--d2l-color-orange-70: #bc2d14;
+			--d2l-color-orange-80: #a82314;
+			--d2l-color-orange-90: #8f1e14;
+		}
+	`;
+	document.head.appendChild(style);
+}
```
dlockhart, line 43: If you like, we're not necessarily tied to the same mechanism we use to inject colour variables in core. You could just define these in a global shared CSS file:

```css
html {
  --d2l-color-gray-a: #fafafa;
  /* etc. */
}
```

The advantage here is that it's one less JS file to import when the page loads, so it should be faster.

dlockhart (approved):

PR #718, opened 2022-02-24
Branding update: component status on page

margaree:
This change componentizes the component status, then uses that on component pages and in the status table. We want it appearing in the top right of the page.


pages/_components/component-status.js
```diff
@@ -0,0 +1,73 @@
+import '@brightspace-ui/core/components/colors/colors.js';
+import '@brightspace-ui/core/components/icons/icon.js';
+import { CC_STATES, DEV_STATES, STABILITY_STATES } from '../../tools/states.js';
+import { css, html, LitElement } from 'lit-element';
+
+export class DaylightComponentStatus extends LitElement {
```
margaree, line 6: I created a component for the status in order to use it both on component pages and within the status table. The logic is copy-pasted from `status-table.js`

pages/_includes/css/style.css
```diff
@@ -85,4 +90,12 @@ blockquote {
 	color: var(--d2l-color-galena);
 	margin-left: 0;
 	padding-left: 1em;
-  }
+}
+
+d2l-daylight-component-status {
```
margaree, line 95: I just absolutely positioned this in order for it to appear in the proper spot in component pages. The other option would be to add it into the `.eleventy.cjs` markdown parsing in the [heading_open rule](https://github.com/BrightspaceUI/documentation/blob/marketing-rebrand/.eleventy.cjs#L134)

dlockhart: Seem OK as long as on smaller screens this position is still correct.

dlockhart (approved):

pages/_includes/css/style.css
margaree: Ah yeah it does some weird things at smaller screens - we'll have to go the other route then

pages/_includes/layouts/component-issue.njk
```diff
@@ -7,31 +7,38 @@ permalink: "components/{{fileName}}/"
...
-{{ content | safe }}
-
-<div class="d2l-prevent-indexing" style="padding-top: 1.5rem;">
-	{%- if repo -%}
-		<div>This is just a summary - see it on GitHub for full details:</div>
-		{% gitHubLink repo %}
+<div class="d2l-daylight-component-page">
+	{%- if eleventyNavigation.official %}
+		<h1 class="d2l-heading-1">
+			<div class="d2l-daylight-h1-container">
```
margaree, line 19: This piece with the container and `d2l-daylight-component-status` that was added in `.eleventy.cjs` also has to be added here because the regular component pages use the `# <page heading 1>` from the README whereas these pages generated from component issues handle generating their own h1

PR #720, opened 2022-02-24
Branding update: h1 divider

margaree:


pages/_includes/css/style.css
```diff
@@ -55,6 +58,12 @@ div.d2l-daylight-content {
 :not(.d2l-daylight-component-page) > .d2l-daylight-h1-section-wrapper + .d2l-daylight-section-wrapper {
 	margin-top: 1.8rem;
 }
+.d2l-daylight-h1-border {
+	border: 3px solid var(--d2l-color-green);
+	margin-bottom: 1.8rem;
+	margin-top: 1.1rem;
+	max-width: 4rem;
+}
```
dlockhart, line 66: From some experiments, I think you can accomplish this without the extra element by using something like this:

```css
.d2l-daylight-content h1.d2l-heading-1::after {
	border: 3px solid var(--d2l-color-green);
	display: block;
	margin-bottom: 1.8rem;
	margin-top: 1.1rem;
	max-width: 4rem;
}
```

You might need to also make the heading a block element.

margaree: Nice! That does seem to work🎉

dlockhart (approved):

PR #723, opened 2022-02-24
Branding: Add additional info box and footer

margaree:


pages/_components/component-owner.js
```diff
@@ -0,0 +1,52 @@
+import '@brightspace-ui/core/components/link/link.js';
+import { css, html, LitElement } from 'lit-element';
+
+const GAUDI_SLACK = '#web-platform';
+
+export class DaylightComponentOwner extends LitElement {
```
margaree, line 6: This was mostly just copied from `status-table.js` in order to use in the additional actions box. It also adds the option to show slack icon, which is used in the additional actions box.

pages/_includes/css/nav.css
```diff
@@ -2,6 +2,7 @@
 	align-self: flex-start;
 	background-color: white;
 	box-shadow: 0 0 7px var(--d2l-color-gypsum);
+	max-height: 100vh;
```
margaree, line 5: Otherwise the side nav would be slightly taller than the page

pages/_includes/css/style.css
```diff
@@ -37,8 +37,12 @@ div.d2l-daylight-content {
 	max-width: 1253px;
 	min-width: 284px;
 	padding: 3.9rem 2.7rem;
+	padding-bottom: 0;
 	width: 100%;
 }
+.d2l-daylight-main-content {
+	min-height: 100vh;
```
margaree, line 43: So the footer will be at the bottom of the display.

There's some weird behaviour happening with the sticky side nav where having the search or really anything that adds any extra spacing is causing additional white space to appear at the bottom of the document. Overflow hidden almost worked but caused problems with the search dropdown positioning.

The compromise here is by having the footer displayed at the very bottom, on pages that don't have enough content to fill the page there becomes some vertical scroll to view the footer. On pages that are longer than page height, this is not an issue, so it's seen in very few cases.

Ideally this `min-height` would be a calculation that would lead to the footer being at the bottom of the visible page, but at the moment that doesn't seem possible.

pages/_includes/layouts/base.njk
```diff
@@ -185,9 +185,17 @@
 				{% sideNav navPages, page.url, toc %}
 			</div>
 			<div class="d2l-daylight-content d2l-body-compact">
-				<div class="d2l-daylight-main-content">
-					{{ content | safe }}
+				<div>
```
margaree, line 188: This extra div is a bit of a hacky way to ensure that the footer is outside the `d2l-daylight-main-content` due to that addition of having section wrappers. I want to revisit that next week.

img/logo-footer.svg
```diff
@@ -0,0 +1,9 @@
...
+  <g data-name="D2L logo">
+    <path data-name="Path 226" d="M0 15.02h1.42l.724-.98V2.34l-.724-.977H0v-.98h7.1c5.452 0 7.865 3.081 7.865 7.652A8.021 8.021 0 0 1 7.1 16H0Zm7.326-.369c3.067 0 4.089-1.945 4.089-5.892S9.782 1.306 6.929 1.306a3.734 3.734 0 0 0-1.164.17l-.4.17V14.04a2.953 2.953 0 0 0 2 .61Z"/>
+    <path data-name="Path 227" d="m25.554 9.143-.8 3.918h-8.513v-.994l1.292-1.417c2.655-2.839 4.756-5.04 4.756-7.34 0-1.42-.568-2.328-1.547-2.328s-1.576.994-1.576 2.484v.752h-2.542a2.967 2.967 0 0 1-.085-.724c0-2.314 2.257-3.492 4.387-3.492s4.259 1.008 4.259 3.407a3.961 3.961 0 0 1-.468 1.775 7.5 7.5 0 0 1-1.079 1.59 23.368 23.368 0 0 1-2.754 2.442c-.838.653-1.576 1.207-1.846 1.42h5.1l1.349-1.42Z"/>
+    <path data-name="Path 228" d="m39.708 11.102-.994 4.9H27.059v-.98h1.462l.738-.98V2.285l-.738-.98h-1.462V.312h7.964v.994h-1.916l-.681.98v12.256h3.691l3.222-3.44Z"/>
+    <path data-name="Path 229" d="M25.086 14.509v.612l-.852.852h-8.433v-.61l.852-.852Z" fill="#34e82a"/>
+  </g>
+</svg>
```
dlockhart, line 9: Very nit: all the `data-name` attributes are artifacts from whatever graphics program was used to create this, so they can be removed.

pages/_components/component-owner.js
```diff
@@ -0,0 +1,52 @@
...
+		super();
+
+		this.showSlackIcon = false;
+	}
+
+	render() {
+		if (!this.owner || this.owner === 'Gaudi' || this.owner.includes('#')) {
+			const slack = (!this.owner || this.owner === 'Gaudi') ? GAUDI_SLACK : this.owner;
+
+			const icon = this.showSlackIcon ? html`<img src="/img/slack-logo.svg" />` : null;
```
dlockhart, line 44:
```suggestion
			const icon = this.showSlackIcon ? html`<img src="/img/slack-logo.svg" alt="" />` : null;
```

pages/_includes/css/style.css
dlockhart: Interesting. This would likely require a bit of CSS-rearchitecture, but I'm wondering if using CSS grid layout for the high-level layout (side nav, content & footer) would help? It's really good for having something positioned all the way left or all the way at the bottom and having other content take up "the rest".

dlockhart (approved):

PR #726, opened 2022-02-25
Search: add f key as shortcut, only open search if focus on body

margaree:
There was a bug where if I was in a text input and typed s, the search would open. I clicked around and having the key only open on `body` as target seems to cover all desired cases, but not sure if there is a more standard approach to this.

We also want the 'f' key to be an option for opening the search so this adds that.


annabelsegalld2l (approved):

dlockhart (approved):
You may be able to also check `e.defaultPrevented` and not do anything if it's `true`. I would have expected the text input to have called `e.preventDefault()` when the character got typed out.

margaree:
The text input does not seem to call `e.preventDefault` so sadly it seems we can't check that. I also tried adding a listener for `@input` to the `input-text` then calling `e.preventDefault()` but the document keydown listener gets triggered first.

dlockhart:
Weird! Ah well, what you have seems to work!

PR #753, opened 2022-03-03
Simplify available layouts for consistency

margaree:
This PR changes many files but most are just updating the layout to use the `simple-content` rather than `base`.

This changes removes any page from using `base`, instead having `base` be the general layout that the other layouts are built off of. This allows us to be able to handle the closing `</div>` that comes from a `div` in the eleventy parsing code under `heading_open`.

This also removes several layouts where another layout could be used, and moves the css as needed.

The file `demo.md` just wasn't needed anymore so I removed that.


.eleventy.cjs
```diff
@@ -22,6 +22,13 @@ module.exports = function(eleventyConfig) {
 	eleventyConfig.addNunjucksShortcode('issue', (issueUrl) => {
 		return `Looking for more details on the component or want to add your input? <d2l-link href="${issueUrl}">Check out the GitHub Issue</d2l-link>`;
 	});
+	eleventyConfig.addNunjucksShortcode('importComponent', (component) => {
```
margaree, line 25: This allows us to import component scripts into an individual markdown/njk file.

pages/_helpers/base-imports.js
```diff
@@ -1,5 +1,6 @@
 import '../_components/component-owner.js';
 import '../_components/component-status.js';
+import '../_components/demo-snippet.js';
```
margaree, line 3: This not being in `base-imports` just caused confusion whenever code snippets weren't appearing properly. Most of our pages use code snippets so might as well just import this universally.

pages/_includes/layouts/base.njk
```diff
@@ -186,10 +186,8 @@
 				{% sideNav navPages, page.url, toc %}
 			</div>
 			<div class="d2l-daylight-content d2l-body-compact">
-				<div>
```
margaree, line 189: Not using this template, and instead using `simple-content` or another wrapper allows us to remove this hacky extra div

pages/_includes/layouts/component-issue.njk
```diff
@@ -1,5 +1,5 @@
 ---
-layout: layouts/demo
+layout: layouts/base
```
margaree, line 2: All layouts are now off of `base` removing extra nesting.

dlockhart (approved):

PR #766, opened 2022-03-09
Search: Include partial matches

margaree:


pages/_components/search.js
```diff
@@ -214,7 +214,7 @@ export class DaylightSearch extends LitElement {
 	}
 
 	_search(searchTerm) {
-		const results = this.index.search(searchTerm);
+		const results = this.index.search(`${searchTerm} *${searchTerm}*`); // match exact term and partial matches
```
margaree, line 217: We need to include both the individual search term and the wildcard because with just the wildcards, exact matches are not included.

This is a potential performance concern. Our index side seems to be small enough that I have not noticed it as an issue, but it would be something to keep an eye on.

dlockhart (approved):

PR #788, opened 2022-03-24
Support hiding content that does not need to be externally facing

margaree:
This adds the ability to hide content that shouldn't be visible externally (e.g., internal links). This does make the assumption that it okay to simply this content and not completely not render it. It is a much more complex change to just not render that content. In general this content would be visible to external users anyways since this repo is public as are any components we currently have in Daylight currently.

This also adds the ability to hide components that are in private repos going forward by adding an "Internal" label to these issues.


.eleventy.cjs
```diff
@@ -116,7 +116,7 @@ module.exports = function(eleventyConfig) {
 		}
 	};
 	const anchorOpts = {
-		level: [2, 3],
+		level: [2, 3, 4],
```
margaree, line 119: I noticed that h4s weren't being anchor-ified, but one was actually being linked to, so I added them

pages/about/adding-component.md
```diff
@@ -139,30 +139,35 @@ Add the comment `<!-- docs: demo -->` directly before the code block. For exampl
 
 #### Hidden Content
 
-This includes information such as installation and testing instructions. Wrap these sections in `<!-- docs: start hidden content -->` and `<!-- docs: end hidden content -->`. For example:
+This includes information such as installation and testing instructions. Wrap these sections in `<!-- docs: start hidden content -->` and `<!-- docs: end hidden content -->` (add a space between start and hidden if copying the example below). For example:
 
 ```
-<!-- docs: start hidden content -->
+<!-- docs: starthidden content -->
```
margaree, line 145: These demos weren't working correctly - the content was actually just being hidden (I think previously we weren't importing the demo-snippet component so that wasn't happening)

pages/about/adding-component.md
```diff
@@ -179,13 +184,17 @@ If you have a component idea that you would like to suggest then you can [submit
 
 ### Labs Component
 
-Labs components generally do not have complete documentation pages in the Daylight site, but do appear in the Status table and also have pages generated from their issues. To document these, [use the GitHub issue template](https://github.com/BrightspaceUI/documentation/issues/new?assignees=&labels=Labs+Component&template=component-in-progress.md&title=%3CComponent+Name%3E).
```
margaree, line 182: The templates no longer exist, so no longer directing to them

pages/designing/design-process.md
```diff
@@ -11,8 +11,12 @@ eleventyNavigation:
...
 Our design process was carefully developed to reflect the phases designers go through from start to finish. We share this process with our counterparts in Product Management and Development.
 
+<!-- docs: start internal content -->
 See [Design Process](https://desire2learn.atlassian.net/wiki/spaces/PD/pages/1850115614/Overview) (on Confluence) for more information.
+<!-- docs: end internal content -->
 
-<a class="d2l-daylight-card" href="https://desire2learn.atlassian.net/wiki/spaces/PD/pages/1850115614/Overview">
-  <img src="/img/design-process-large.png" alt="Design process diagram showing 7 hexagons labelled with the design phases: Understanding, Empathizing, Exploring, Defining & Crafting, Testing & Iterating, Refining & Extending, Executing">
-</a>
+<div class="d2l-daylight-internal-content">
```
dlockhart, line 18: Could the `class` go directly on the `<a>`? Or alternatively could the whole thing be moved inside the `<!--` block right above it?

tools/build-index.js
```diff
@@ -186,6 +186,7 @@ function readHtml(filename) {
 
 	// remove content that shouldn't be indexed
 	$('.d2l-prevent-indexing').remove();
+	$('.d2l-daylight-external .d2l-daylight-internal-content').remove();
```
dlockhart, line 189: Neat, that's super elegant!

dlockhart (approved):
Very cool, and much easier to do than I would have thought!

package.json
```diff
@@ -5,7 +5,10 @@
   "scripts": {
     "build": "npm run clean && node ./tools/generate-pages.js && npm run build:rollup:wc && npm run build:rollup",
     "build:index": "node ./tools/build-index.js",
-    "build:prod": "cross-env NODE_ENV=production npm run build && cross-env NODE_ENV=production npx eleventy --config=.eleventy.cjs && cp -r assets _site/",
+    "build:prod": "npm run build:prod:build && npm run build:prod:eleventy",
+    "build:prod:external": "cross-env AUDIENCE=external npm run build:prod:build && cross-env AUDIENCE=external npm run build:prod:eleventy",
+    "build:prod:build": "cross-env NODE_ENV=production npm run build",
+    "build:prod:eleventy": "cross-env NODE_ENV=production npx eleventy --config=.eleventy.cjs && cp -r assets _site/",
```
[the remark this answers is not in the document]
dlockhart: So is the idea that the different sites will get deployed to different S3 buckets and then we'll somehow serve up different buckets based on the user's IP address?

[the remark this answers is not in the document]
margaree: My original thought had been yes to deploying to different S3 buckets but then there would be a separate URL for external users which wouldn't have authentication. If we could do the serving up different buckets based on IP address then that seems more ideal.

pages/designing/design-process.md
margaree: The adding the class to the a should work. There's some issue with moving it inside the `<!--` that would require some rework but I'll dig a bit into that.

PR #795, opened 2022-03-25
Backlog component page

margaree:
- Shows Rally link if available and internally facing
- Adds Color Input backlog component


pages/_includes/css/page-content.css
```diff
@@ -63,26 +63,33 @@
...
 }
+
 .d2l-github-logo {
 	background-color: var(--d2l-color-celestine);
 	padding: 0.5rem;
 }
+.d2l-rally-logo {
+	background-color: white;
+	border-right: 1px solid var(--d2l-color-gypsum);
+	height: 33px;
```
dlockhart, line 90: Nit: you may want to resize the actual image -- which is around `200px` square -- to be this size. That way the file is smaller and the browser doesn't have to resize it at runtime.

dlockhart (approved):
Cool! 😎

PR #812, opened 2022-03-29
Modifications to development state logic

margaree:
- Modify states code to prioritize deprecated and backlog labels over development state in issue comment
- Fix to calculate development state after info has actually been processed


test/utils/generate-pages-utils.test.js
```diff
@@ -20,7 +20,7 @@ describe('generate-pages', () => {
 			{ labels: [], state: 'closed', issueState: 'Backlog', expected: 'Backlog' },
 			{ labels: [], state: 'closed', issueState: 'Not Started', expected: 'Not Started' },
 			{ labels: [{ name: 'Backlog' }], state: 'open', issueState: 'In Progress', expected: 'Backlog' },
-			{ labels: [{ name: 'Deprecated' }], state: 'open', issueState: 'In Progress', expected: 'In Progress' },
+			{ labels: [{ name: 'Deprecated' }], state: 'open', issueState: 'In Progress', expected: 'Deprecated' },
```
margaree, line 23: This would in theory never happen, but in practice it makes more sense to prioritize the Deprecated label

tools/generate-pages-utils.js
```diff
@@ -4,17 +4,14 @@ import matter from 'gray-matter';
...
-		else if (Object.values(DEV_STATES).includes(issueState)) return issueState;
-		else return DEV_STATES.COMPLETE;
-	} else if (state === 'open') {
-		if (labelNames.includes(ISSUE_LABELS.BACKLOG)) return DEV_STATES.BACKLOG;
-		else if (Object.values(DEV_STATES).includes(issueState)) return issueState;
-		else return DEV_STATES.NOT_STARTED;
-	}
-	return 'Unknown'; // in the very unlikely event the issue doesn't have a state of open or closed
+
+	if (labelNames.includes(ISSUE_LABELS.DEPRECATED)) return DEV_STATES.DEPRECATED;
```
margaree, line 8: Prioritizing the label over the development state that's hidden in the issue comment makes more sense as these are clearer to communicate to a casual viewer of the issue. i.e., Backlog and Deprecated labels on the issue make that piece obvious and easy to change. We might also want to just make "In Progress" also be a label and then that would be clear as well and it wouldn't have to be hidden away.

dlockhart (approved):

PR #888, opened 2022-06-09
Add documentation around updating component state and upgrading component tier, other minor tweaks

margaree:
This PR does the following:
- minor tweaks found when using new custom-elements.json
- Adds documentation related to editing component pages
- Updates the state process to primarily rely on issue labels rather than sometimes using issue labels and sometimes using hidden issue comment content (continues to support the old way while issues are being converted over)


pages/_components/demo-tables.js
```diff
@@ -153,7 +153,7 @@ export class DaylightDemoTables extends ScrollShadowRootMixin(LitElement) {
 
 		if (this._componentInfo && this._componentInfo.attributes) {
 			this._componentInfo.attributes.forEach((attribute) => {
-				if (attribute.description.startsWith(ADVANCED)) {
+				if (attribute.description && attribute.description.startsWith(ADVANCED)) {
```
margaree, line 156: When testing with the new analyzer custom-elements.json files, ran into these two cases that causes errors

pages/about/editing-page.md
```diff
@@ -70,3 +70,70 @@ This method likely isn't needed, but in the event you want to hide specific HTML
...
+- `Inactive`: Add the `Inactive` label to the issue
+- `On the backlog`: Add the `Backlog` label to the issue
+- `Not Planned`: Mark the issue as opened
+- `Completed`: Mark the issue as closed
+
+For each of the above cases, remove any other state label.
+
+**Deprecated**
+
+You may see the development state indicated as `development: Inactive | In Progress` in the hidden content throughout GitHub issues. Please use the label process from above going forward.
```
margaree, line 100: I'm changing the process here because it was super mixed (e.g., labels for some states, hidden content for other states)

pages/_components/demo-tables.js
dlockhart: Likely safe to use optional chaining here alternatively: `attribute.description?.startsWith(ADVANCED)`

dlockhart (approved):

PR #896, opened 2022-06-13
Remove support for including stability and development states in component issue hidden content

margaree:
Also add note about rate limit error

All component issues have been updated to no longer include this information.


pages/about/editing-page.md
```diff
@@ -77,11 +77,7 @@ Component stability and development statuses are generally modified using GitHub
 
 ### Stability
 
-Add the `Stable` label if the component is `Stable`. Add the `Not Stable` label if the component is `Not Stable`.
-
-**Deprecated**
-
-You may see the stability state indicated as `stability: Stable | Not Stable` in the hidden content throughout GitHub issues. Please use the label process from above going forward.
+If the component has a development stated of `Completed` and is `Stable`, closing the issue by default will give it these states. Otherwise, add the `Stable` label if the component is `Stable`, or add the `Not Stable` label if the component is `Not Stable`.
```
dlockhart, line 80:
```suggestion
If the component has a development state of `Completed` and is `Stable`, closing the issue by default will give it these states. Otherwise, add the `Stable` label if the component is `Stable`, or add the `Not Stable` label if the component is `Not Stable`.
```

dlockhart (approved):

PR #911, opened 2022-06-23
Notify slack on failed deploy

svanherk:
Not sure if there's an easy way to make a deploy fail to test this.


margaree (approved):
Yup so one way to test a deployment failure is remove a demo component from the array in the component issue (e.g., tooltip-help from the tooltip issue), then the build should fail during the deployment.

svanherk:
So I removed one from `hierarchical-view` and it didn't fail the deploy, but the demo page definitely broke.  I broke the route to the README though and that worked!  Both fixed now

PR #913, opened 2022-06-24
Use new token for issue-label-change.yml

svanherk:
This uses our new write token, since `secrets.GITHUB_TOKEN` doesn't trigger additional actions (in this case CI).


dlockhart (approved):
Nice! So this will mean that the automated PRs from label changes will now get a site deployed!

PR #915, opened 2022-06-24
Fix token passing

svanherk:
I _think_ this is what I need to do for GitHub app- tokens, based on https://github.com/Brightspace/hello-world-fra-iframe/blob/main/.github/workflows/update-package-lock.yml 🤞


dlockhart (approved):
yolo

PR #622, opened 2022-02-04
Search: Simplify index generation, prep for linking with frontend

margaree:
This PR is to prepare the index building code to contain what we want for showing on the frontend. I did a decent amount of refactoring to try to simplify a bit. Apologies for the giant diff - it's probably easier to comprehend reading the files themselves rather than the diff (and I'll add some comments to try to assist).

PR #918, opened 2022-06-27
Search: enable searching on properties, slots, and events

margaree:
Properties and slots were never searchable previously but had been implemented. During testing when implementing search problems were discovered (see https://github.com/BrightspaceUI/documentation/pull/622#discussion_r799624960) so it was disabled.

This PR enables searching on properties, slots and events (adding in events as a new type). It also includes bolding the property/event/slot name in the search result.


pages/_components/search-result.js
```diff
@@ -166,14 +166,19 @@ class SearchResultMenuItem extends MenuItemMixin(LitElement) {
...
 			onlyParentHeader = false;
 			topLevelPage = this._item.sectionHeader || this._item.parentHeader;
 			secondarySection = RECORD_TYPE_CAT_NAME[this._item.type];
 			if (typeof(this._item.preview) === 'object') {
 				const supportingName = this._item.preview.name || this._item.preview.attribute || (this._item.type === RECORD_TYPE.SLOT ? 'Default' : '');
 				const supportingDescription = `${this._item.preview.description && supportingName ? ': ' : ''}${this._item.preview.description || ''}`;
-				supportingContent = html`${supportingName ? html`<code>${supportingName}</code>` : ''}${supportingDescription}`;
+				const name = supportingName ? html`<code>${unsafeHTML(boldSearchTerm(supportingName, this.searchTerm))}</code>` : '';
+				const description = unsafeHTML(boldSearchTerm(supportingDescription, this.searchTerm, true));
+				supportingContent = html`${name}${description}`;
```
margaree, line 181: Bold the search term if in one of these results

pages/_components/search.js
```diff
@@ -215,7 +215,7 @@ export class DaylightSearch extends LitElement {
 	}
 
 	_search(searchTerm) {
-		const results = this.index.search(`${searchTerm} *${searchTerm}*`); // match exact term and partial matches
+		const results = this.index.search(`${searchTerm}^50 *${searchTerm}*`); // match exact term and partial matches
```
margaree, line 218: This boosts an exact match to the top of the results. Note that if this is a multi-word query then this boosts the last word and it doesn't seem like lunr has implemented multi-word exact query searching. In practice/in my anecdotal experience I don't think this is a huge deal (e.g., most of my queries are single words or the query still returns the result I'm looking for near the top) but it's not totally ideal.

tools/build-index.js
```diff
@@ -44,11 +44,15 @@ function buildDemoRecords(tagName, url, pageTitle, sidenavSection, sectionTitle)
 		return buildRecordPreviewFromObject(url, pageTitle, sidenavSection, property, sectionTitle, RECORD_TYPE.PROPERTY);
 	});
 
+	const eventRecords = (componentInfo.events || []).map((property) => {
```
margaree, line 47: Events weren't being included in search results even when `SHOW_PROPERTIES` was enabled, so I implemented that

tools/build-index.js
```diff
@@ -59,7 +63,12 @@ function buildIndex(docs) {
 		this.field('c'); // content
 		this.field('u'); // url
 		docs.forEach(function(doc) {
-			this.add(doc);
+			if (doc.t !== RECORD_TYPE.PROPERTY
+				&& doc.t !== RECORD_TYPE.EVENT
+				&& doc.t !== RECORD_TYPE.SLOT
+				&& doc.t !== RECORD_TYPE.A11Y_PROPERTY) {
+				this.add(doc, { boost: 2 });
```
margaree, line 70: This boosts the importance of docs that aren't properties/events/slots since the previous problem with showing the properties results was that they would appear ahead of regular page results and dominate the results (e.g., searching button would first show all its properties before other button pages)

pages/_components/search.js
dcostiuc: What exactly does the `^50` do/why does it boost an exact match? 
It is because it's then more likely to favour results with that exact term?

How come you chose `50` and not another number?

pages/_components/search-result.js
dcostiuc: Just double-checking, but with the use of `unsafeHTML`, are there any security concerns there? 
This should just be used by internal employees, but could an internal bad actor still do anything with a specially-formatted search string?

tools/search-utils.js
```diff
@@ -7,11 +7,13 @@ export const RECORD_TYPE = {
 	A11Y_PROPERTY: 'cap',
 	CONTENT: 'cc',
 	DEMO: 'cd',
+	EVENT: 'ce',
```
dcostiuc, line 10: Just wondering about the naming convention. 
What does the `c` in all of these record type values represent?

dcostiuc: Oh nevermind, I think that's explained [here](https://github.com/BrightspaceUI/documentation/pull/918/files#diff-430b2b3912e72269b06307c42c0d5be39b7a4e427d73a28f1bfb659e972a9bb6L87).

tools/search-utils.js
```diff
@@ -62,7 +64,7 @@ export function generatePreview(record, sidenavSection, type, preview) {
 	};
```
dcostiuc, line 64: So [here ](https://github.com/BrightspaceUI/documentation/pull/918/files#diff-430b2b3912e72269b06307c42c0d5be39b7a4e427d73a28f1bfb659e972a9bb6R62)we already have some code support for `type`. 

Just wanted to confirm, but the context for this PR is that we previously had a lot of this support (for searching properties and slots), but noticed some problems with it, so a lot of it was disabled/removed, and now you're adding it back in, plus event support, right?

tools/build-index.js
```diff
@@ -34,7 +34,7 @@ const EXCLUDE_FILES = [
 const HTML_FOLDER = './_site';
 const OUTPUT_INDEX = '_site/lunr_index.json';
 const OUTPUT_PREVIEW = '_site/search_preview.json';
-const SHOW_PROPERTIES = false;
+const SHOW_PROPERTIES = true;
 
 // PROPERTIES & SLOTS: This function currently isn't used but will be if properties & slots code in parseComponentPage is enabled
```
dcostiuc, line 39: Should this comment be updated/removed? Where is `parseComponentPage` defined? (couldn't find it in this repo)

tools/build-index.js
dcostiuc: So how does `boost` work? Why boost `2` here and not another number?

tools/build-index.js
margaree: Yup will update. That should be `parsePage` - I believe it previously was `parseComponentPage` but was refactored to be more general

tools/search-utils.js
```diff
@@ -39,7 +41,7 @@ export function buildRecordPreviewFromObject(url, pageTitle, sidenavSection, inf
 
 	const content = `${infoObject.name} ${infoObject.type || ''} ${infoObject.description || ''}`;
 
-	const record = generateRecord(url, pageTitle, sectionTitle, content, id, urlSuffix);
+	const record = generateRecord(url, pageTitle, sectionTitle, content, id, urlSuffix, type);
```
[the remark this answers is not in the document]
margaree: Not currently. It actually should be removed before it's used in generating the index in `buildIndex` so I'll update that as well.

An alternative is to go the route of only sending the type in the generic content case (https://github.com/BrightspaceUI/documentation/blob/main/tools/build-index.js#L163) and checking for that in buildIndex. Doing it this way felt more explicit and safe in case more record types were added.

dcostiuc (commented):
Played around with this PR's [Preview URL](https://pr-918-daylight.d2l.dev/about/overview/) and looks like it mostly works as expected!
(can see search results pop up for properties, slots, and they seem to be lower in the list compared to other results)

Left some comments/questions for consideration.

However, worth noting that I didn't seem to see any expected search results for Tooltip's [events](https://pr-918-daylight.d2l.dev/components/tooltip/#d2l-tooltip-events), e.g. whether searching for `event` or `d2l-tooltip-show` or `Dispatched when the tooltip is opened`

margaree:
> Played around with this PR's [Preview URL](https://pr-918-daylight.d2l.dev/about/overview/) and looks like it mostly works as expected! (can see search results pop up for properties, slots, and they seem to be lower in the list compared to other results)
> 
> Left some comments/questions for consideration.
> 
> However, worth noting that I didn't seem to see any expected search results for Tooltip's [events](https://pr-918-daylight.d2l.dev/components/tooltip/#d2l-tooltip-events), e.g. whether searching for `event` or `d2l-tooltip-show` or `Dispatched when the tooltip is opened`

Yup that's related to [this](https://github.com/BrightspaceUI/documentation/pull/918#discussion_r907535661) issue with being unable to search multi word phrases (hyphenation seems to be processed the same). So while this PR does add a lot of properties/slots/events searching and is an improvement it isn't perfect. I'll play around a bit more but so far haven't gotten anything to work/based on reading some lunr issues multi-word search just doesn't exist.

pages/_components/search-result.js
margaree: Based on [the documentation](https://lit.dev/docs/templates/directives/#unsafehtml) we want its usages to be developer controlled. In theory yes an internal bad actor could potentially get something bad through, but in terms of these cases they would have to modify the name or description of a component in a malicious way, which would then have to pass code review, and so to me this is unlikely.

I'll take a look into your idea Erik, thanks.

tools/search-utils.js
dcostiuc: Huh, looks like Github is having more than 1 kind of problem today perhaps haha.

With that link, I just meant this snippet

```
return {
		...record,
		s: sidenavSection,
		**t: type,**
```

dcostiuc: > Properties and slots were never searchable previously but had been implemented. During testing when implementing search problems were discovered (see [#622 (comment)](https://github.com/BrightspaceUI/documentation/pull/622#discussion_r799624960)) so it was disabled. Otherwise yup that's correct.

May be worth mentioning in the PR description for future reference/context. But that's just an optional idea/suggestion

margaree: Oh no, poor GitHub!

Yup you're right, preview does already include type. `buildIndex` specifically includes the records and not previews, but since the preview adds in the record we should actually not include that a second time in `generatePreview`. Will update.

Good idea - will update the description as well. For some reason I thought I'd included that somewhere and definitely didn't🤦‍♀️

tools/search-utils.js
dcostiuc: If we think that we're not going to have `mc` or some other type that doesn't begin with a non-c in the future, then yeah I would agree that we could remove the c if we wanted to. Otherwise might not be worth bothering.

tools/search-utils.js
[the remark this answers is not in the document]
dcostiuc: > It actually should be removed before it's used in generating the index in buildIndex so I'll update that as well. (nevermind - looks like it's not included in the index because it's not one of the fields set above)

Sorry, not sure what you're referring to there.

> An alternative is to go the route of only sending the type in the generic content case (https://github.com/BrightspaceUI/documentation/blob/main/tools/build-index.js#L163) and checking for that in buildIndex. Doing it this way felt more explicit and safe in case more record types were added.

So is this what we currently do? If so, it sounds like the way to go!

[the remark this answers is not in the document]
margaree: I was referring to the `type` - I had been concerned that it was being unnecessarily added to the lunr_index file that gets generated but that's not actually the case.

Yup, this code change is currently using the more explicit method of checking the type in `buildIndex`👍

margaree:
I have something almost ready to go on this. I'm going to open a fresh PR though since there were fairly significant changes and I don't want this to be super confusing (I'll cross-mention comments here though as needed)

PR #927, opened 2022-07-04
Add Rally feature information to component request issue template

margaree:
Used here https://github.com/BrightspaceUI/documentation/blob/main/pages/_includes/layouts/component-issue.njk#L49 but was not yet documented.


dcostiuc:
So what does this change do? 

Is the idea that someone creating a new component request issue could also now more easily fill in some Rally info and uncomment what was added in Files changed? 

Would that also cause any other kind of integration related to the documentation site to go off?

Also, you may want to merge `main` into this branch (seems to be out-of-date).

margaree:
This adds the comment block at the top to the issue template so when someone is adding a new component request they are also easily able to add the Rally feature block (can be seen for example at the bottom of this page https://daylight.d2l.dev/components/color-input-&-color-contrast-analyzer/)

They would not uncomment the block. It would be left as is and filled in where specified. This pattern can be seen across nearly all our component issues if you click the "..." > "Edit".

There are no other changes. This is purely to document an existing functionality.

dlockhart (approved):

PR #937, opened 2022-07-07
Search: Enable Properties work PR 2: Enable searching hyphenated words

margaree:
It turns out there is a way to allow searching for hyphenated words and get exact matches.

The current lunr term separate is both whitespace and a hyphen (see `lunr.tokenizer.separator` in https://lunrjs.com/docs/tokenizer.js.html). This changes it to just be the whitespace.

To see an example of this, search `inherit-font-style` on the PR site vs on production Daylight.

One note is that the previous behaviour would search all the words (inherit, font, style). There is a way to implement this as well so that it would search `[inherit-font-style, inherit, font, style]`. This does complicate the results a bit and I'm not sure if this is necessary - e.g., how often someone would be searching for a particular hyphenated word like a property or tag and want to see all possible results with the containing words as well. Let me know any thoughts there.


eKoopmans (approved):
Looks good, except for the weird visual-diff failure...

PR #941, opened 2022-07-08
Search: Enable Properties Part 3: Remove unsafeHTML from search-result

margaree:
Based on suggestion here https://github.com/BrightspaceUI/documentation/pull/918#discussion_r907533611

The use of HTML within the search results added quite a bit of complexity and was in a security grey area. It was only really used for `<code>` tags and its usefulness was arguable. 

~All code is stripped out at the build index step now (previously all but `<code>` was).~

This still bolds the search term but otherwise leaves the text as text. There will be changes to properties/events/slots results as well related to this, but that will be done once they're actually enabled.


pages/_components/search-result.js
```diff
@@ -4,25 +4,29 @@ import { COMPONENT_SUB_TYPE, MAX_PREVIEW_CHARS, RECORD_TYPE, RECORD_TYPE_CAT_NAM
...
-export function boldSearchTerm(text, searchTerm, replaceLtGt) {
+export function boldSearchTerm(text, searchTerm) {
 	if (!text || typeof(text) === 'object') return text;
-	if (replaceLtGt) text = text.replace(/</g, '&lt;').replace(/>/g, '&gt;');
 	if (!searchTerm) return text;
 
 	const re = new RegExp(`(${searchTerm})`, 'gi');
-	return text.replace(re, '<span class="d2l-search-result-term">$1</span>');
+	const splitString = text.split(re);
+	let accumulator = html``;
```
margaree, line 14: Based on Erik's suggestion https://github.com/BrightspaceUI/documentation/pull/918#discussion_r907644101

pages/_components/search-result.js
```diff
@@ -4,25 +4,29 @@ import { COMPONENT_SUB_TYPE, MAX_PREVIEW_CHARS, RECORD_TYPE, RECORD_TYPE_CAT_NAM
...
+	splitString.forEach(partialString => {
+		if (partialString.toLowerCase() === searchTerm.toLowerCase()) accumulator = html`${accumulator}<span class="d2l-search-result-term">${partialString}</span>`;
+		else accumulator = html`${accumulator}${partialString}`;
+	});
+	return accumulator;
 }
 
 export function getSupportingContent(content, searchTerm, startIndex) {
```
margaree, line 22: This will now return `resultHTML`, since that's what is returned from `boldSearchTerm` and `resultText` to be used in the `aria-description` for the result.

pages/_components/search-result.js
```diff
@@ -164,28 +166,23 @@ class SearchResultMenuItem extends MenuItemMixin(LitElement) {
...
 		let secondarySection = '';
-		let supportingContent = getSupportingContent(this._item.preview, this.searchTerm, 0) || '';
+		const supportingContent = getSupportingContent(this._item.preview, this.searchTerm, 0);
 		let onlyParentHeader = !this._item.sectionHeader || this._item.sectionHeader === this._item.parentHeader; // whether to only show main page header
 
 		if (this._item.type === RECORD_TYPE.PROPERTY || this._item.type === RECORD_TYPE.SLOT || this._item.type === RECORD_TYPE.A11Y_PROPERTY) {
 			onlyParentHeader = false;
 			topLevelPage = this._item.sectionHeader || this._item.parentHeader;
 			secondarySection = RECORD_TYPE_CAT_NAME[this._item.type];
-			if (typeof(this._item.preview) === 'object') {
```
margaree, line 174: These cases are no longer going to have previews that are objects (will change in `build-index` when adding properties/events/slots to the search results). Having 2 different possible return types is complicated and we can just generate the preview when building the index rather than here.

tools/build-index-utils.js
```diff
@@ -30,7 +30,7 @@ export function generatePreview(record, sidenavSection, type, preview) {
 		...record,
 		s: sidenavSection,
 		t: type,
-		p: preview || record.c
+		p: preview || ''
```
margaree, line 33: This was causing keywords to show up in results where there wasn't preview text (e.g., Collapsible Containers) and there isn't a usecase I could find to keep it.

tools/build-index.js
```diff
@@ -125,7 +125,7 @@ function parsePage(url, sidenavSection, $) {
 
 	let keywords = $('.d2l-search-keywords');
 	keywords.remove();
-	if (keywords) keywords = keywords.text().replace(',', ' ');
+	if (keywords) keywords = keywords.text().replace(/,/g, ' ');
```
margaree, line 128: Only the first comma was being removed

tools/build-index.js
```diff
@@ -155,15 +155,13 @@ function parsePage(url, sidenavSection, $) {
...
 				if (!text) return;
-				let textHtml = $(el).html();
-				textHtml = textHtml.replace(/(\r\n|\n|\r)/gm, ' ').replace(/<(?!\/?code)[^>]*>/gm, ''); // filter out all tags except <code></code>
 				contentText += ` ${text}`;
-				contentHtml += ` ${textHtml}`;
+				previewText += ` ${text}`;
 			}
 		});
 
 		const record = generateRecord(url, pageTitle, title, contentText);
```
margaree, line 163: `contentText` is what the actual search uses to search on and includes the keywords. `previewText` is what is shown in the search results and we don't want to contain the keywords.

test/components/search-result.test.js
```diff
@@ -32,64 +32,109 @@ describe('d2l-daylight-search-result', () => {
...
 		it('should not bold anything if no search term present', () => {
 			expect(boldSearchTerm('my input')).to.equal('my input');
 		});
 
-		it('should bold search term', () => {
-			expect(boldSearchTerm('my input', 'input')).to.equal('my <span class="d2l-search-result-term">input</span>');
+		it('should bold search term', async() => {
+			const res = boldSearchTerm('my input', 'input');
+			const resFixture = await fixture(html`<div>${res}</div>`);
+			expect(resFixture).dom.to.equal('<div>my <span class="d2l-search-result-term">input</span></div>');
```
margaree, line 47: To use `.dom` it seems that the res has to be wrapped in a `div`, otherwise it just returns the dom starting with the first tag (i.e., the `span`)

pages/_components/search-result.js
oelhanafey: Is it possible to use an accumulator like this in a similar way to maintain the `<code>` block formatting in the descriptions?

test/components/search-result.test.js
oelhanafey: I'm a bit curious why you decided to change the tests in this way. Does rendering the html and then comparing the result test anything different than just comparing the strings like you were doing before?

oelhanafey (commented):
It looks good to me but I will let someone else have a look and approve it!

test/components/search-result.test.js
margaree: Yup so previously `boldSearchTerm` and `getSupportingContent` returned text and then we later used `unsafeHTML` to render that text, so a direct comparison was possible. Now `boldSearchTerm` and `getSupportingContent` (`resultHTML`) return html and so chai doesn't deal well with a direct comparison (e.g., ` AssertionError: expected { Object (_$litType$, strings, ...) } to equal 'my <span class="d2l-search-result-term">input</span>'`)

pages/_components/search-result.js
margaree: Ah yup it likely is - I'll give it a try and see if it's fairly straightforward

test/components/search-result.test.js
oelhanafey: Oh, makes sense. Thanks!

tools/build-index-utils.js
dcostiuc: You mean like `"show hide,more,less"` showing up in the search result when searching for `Collapsible Containers` in the currently-deployed documentation site? And that being unnecessary/not making sense, right?

pages/_components/search-result.js
```diff
@@ -4,25 +4,29 @@ import { COMPONENT_SUB_TYPE, MAX_PREVIEW_CHARS, RECORD_TYPE, RECORD_TYPE_CAT_NAM
 import { css, html, LitElement } from 'lit-element/lit-element.js';
 import { MenuItemMixin } from '@brightspace-ui/core/components/menu/menu-item-mixin.js';
 import { menuItemStyles } from '@brightspace-ui/core/components/menu/menu-item-styles.js';
-import { unsafeHTML } from 'lit-html/directives/unsafe-html.js';
```
dcostiuc, line 7: While we're at it in terms of replacing the use of `unsafeHTML`, just double-checking, but is there any concern with the remaining uses of it [here](https://github.com/BrightspaceUI/documentation/blob/search-properties-pr3/pages/_components/demo-snippet.js#L209-L214)?

margaree: No concern, that code is entirely developer controlled via what's coming from the READMEs. It would have to involve a malicious developer and reviewer (or potentially just malicious admin). It would also get picked up on the documentation repo visual diffs. So in theory if someone on Gaudi went rogue then yes something could go wrong but I see it as very unlikely.

pages/_components/search-result.js
eKoopmans: Small suggestion, rather than `let ...` this could use `.reduce`:

```js
	return splitString.reduce((accumulator, partialString) => {
		if (partialString.toLowerCase() === searchTerm.toLowerCase()) return html`${accumulator}<span class="d2l-search-result-term">${partialString}</span>`;
		return html`${accumulator}${partialString}`;
	}, html``);
```

pages/_components/search-result.js
```diff
@@ -174,18 +176,21 @@ class SearchResultMenuItem extends MenuItemMixin(LitElement) {
 			if (typeof(this._item.preview) === 'object') {
 				const supportingName = this._item.preview.name || this._item.preview.attribute || (this._item.type === RECORD_TYPE.SLOT ? 'Default' : '');
 				const supportingDescription = `${this._item.preview.description && supportingName ? ': ' : ''}${this._item.preview.description || ''}`;
-				supportingContent = html`${supportingName ? html`<code>${supportingName}</code>` : ''}${supportingDescription}`;
+				supportingContent = {
+					resultHTML: html`${supportingName ? html`<code>${supportingName}</code>` : ''}${supportingDescription}`,
+					resultText: `${supportingName ? supportingName : ''}: ${supportingDescription}`
```
eKoopmans, line 181: Nit: could use `supportingName ?? ''`.

eKoopmans (approved):
Awesome, thanks for implementing this! I'm a bit unsure about the changes to `code` blocks (commented on the v-diff PR), but the rest looks good 👍.

pages/_components/search-result.js
margaree: So this is possible, the one issue is that if the search term is within the code, bolding it is difficult/potentially not possible (let me know if you have ideas I'm missing!).

For example, say we have `<code>d2l-tooltip</code>` where tooltip is the search term. If we try to have something like
```
accumulator = html`${accumulator}<code>`
```
with the intention of then adding the code with the bold span wrappers after, the `<code>` tag will actually just get auto-closed, so we end up with `<code></code>d2l-<span>tooltip</span>`.

We also can't just use regex to replace with html tags wrapping the search term because then they just get shown as strings.

My concern would be if the search term is bolded in regular text but not in the code text, it does look a bit broken.

pages/_components/search-result.js
```diff
@@ -4,25 +4,29 @@ import { COMPONENT_SUB_TYPE, MAX_PREVIEW_CHARS, RECORD_TYPE, RECORD_TYPE_CAT_NAM
...
 
 	const re = new RegExp(`(${searchTerm})`, 'gi');
-	return text.replace(re, '<span class="d2l-search-result-term">$1</span>');
+	const splitString = text.split(re);
+	let accumulator = html``;
+	splitString.forEach(partialString => {
+		if (partialString.toLowerCase() === searchTerm.toLowerCase()) accumulator = html`${accumulator}<span class="d2l-search-result-term">${partialString}</span>`;
+		else accumulator = html`${accumulator}${partialString}`;
+	});
+	return accumulator;
```
dcostiuc, line 19: Thoughts on using `reduce()` here?

pages/_components/search-result.js
```diff
@@ -4,25 +4,29 @@ import { COMPONENT_SUB_TYPE, MAX_PREVIEW_CHARS, RECORD_TYPE, RECORD_TYPE_CAT_NAM
...
 	if (!text || typeof(text) === 'object') return text;
-	if (replaceLtGt) text = text.replace(/</g, '&lt;').replace(/>/g, '&gt;');
 	if (!searchTerm) return text;
 
 	const re = new RegExp(`(${searchTerm})`, 'gi');
-	return text.replace(re, '<span class="d2l-search-result-term">$1</span>');
+	const splitString = text.split(re);
+	let accumulator = html``;
+	splitString.forEach(partialString => {
+		if (partialString.toLowerCase() === searchTerm.toLowerCase()) accumulator = html`${accumulator}<span class="d2l-search-result-term">${partialString}</span>`;
```
dcostiuc, line 16: This line seems a bit long (need to scroll in Github/VS Code to view it all). Perhaps it may be better to write it out over several lines?

e.g.

```js
if (partialString.toLowerCase() === searchTerm.toLowerCase()) {
   accumulator = html`${accumulator}<span class="d2l-search-result-term">${partialString}</span>`;
}
else {
...
}
```

pages/_components/search-result.js
dcostiuc: Oh nevermind, looks like Erik already [brought that up](https://github.com/BrightspaceUI/documentation/pull/941#discussion_r918264787)

pages/_components/search-result.js
margaree: Yes, keeping it for the moment since its removal isn't really related to the work here and there would also be changes needed to the visual diffs

pages/_components/search-result.js
```diff
@@ -198,7 +203,7 @@ class SearchResultMenuItem extends MenuItemMixin(LitElement) {
 	_setAriaAttributes(topLevelPage, secondarySection, supportingContent) {
 		this.text = secondarySection || topLevelPage;
 		let ariaDescription = `Results category: ${this._item.sidenavSection}. ${topLevelPage} page.`;
-		if (supportingContent && typeof(supportingContent) === 'string') ariaDescription += ` ${supportingContent.replace(/<\/?[^>]*>/gm, '').replace(/&gt;/gm, '').replace(/&lt;/gm, '')}`;
+		if (supportingContent && typeof(supportingContent) === 'string') ariaDescription += ` Result preview: ${supportingContent}`;
```
dcostiuc, line 206: Looks like we're making this clearer for our screen reader users 👍

pages/_components/search-result.js
eKoopmans: @margaree I see the problem. It's not clean, but if you like I could put together a solution that handles both `<code>` and `searchTerm` - my thinking is:

1. In the `.split` regex, search for both things that would require custom parsing:
    ```const re = new RegExp(`(<code>.*?</code>|${searchTerm})`, 'gi');```
2. In the accumulator logic:
    - if it matches the searchTerm, do what's currently done
    - if it matches `<code>.*?</code>`, have a nested accumulator that looks for and splits out `searchTerm`, and sticks it all in ``html`<code>...</code>`
    - otherwise, treat it as regular text

eKoopmans: ☝️ Here's what I've got:

```js
// applyCodeAndSearch would replace boldSearchTerm:
const applyCodeAndSearch(text, searchTerm) {
	if (!text || typeof(text) === 'object') return text;
	const re = new RegExp('(<code>.+?</code>)', 'gi');
	const splitString = text.split(re);

	return splitString.reduce((accumulator, partialString) => {
		const contents = applySearch(partialString, searchTerm);
		const parsed = re.test(partialString) ? html`<code>${contents}</code>` : contents;
		return html`${accumulator}${parsed}`;
	}, ``);
}

const applySearch(text, searchTerm) {
	if (!searchTerm) return text;
	const re = new RegExp(`(${searchTerm})`, 'gi');
	const splitString = text.split(re);

	const searchLower = searchTerm.toLowerCase();
	return splitString.reduce((accumulator, partialString) => {
		const parsed = re.test(partialString) ? html`<span class="d2l-search-result-term">${partialString}</span>` : partialString;
		return html`${accumulator}${parsed}`;
	}, ``);
}
```

PR #947, opened 2022-07-11
Added .nvmrc

oelhanafey:
Added .nvmrc to the repo to make the use of node 14 more clear.


svanherk:
Nice! We should update all the workflows to use the file as well, so the switch to 16 can happen in one place.
For example, switching https://github.com/BrightspaceUI/documentation/blob/main/.github/workflows/ci.yml#L12 to `node-version-file: .nvmrc` like `core` uses

oelhanafey:
Yep, good idea Stacey! Thanks!

svanherk (approved):

PR #949, opened 2022-07-12
Change code block padding to 1.2rem

oelhanafey:
[Rally Defect DE49091](https://rally1.rallydev.com/#/?detail=/defect/641659490621&fdp=true)

Changed the code block padding to a uniform 1.2rem on all sides. The extra padding was intended to be used for buttons at the top of the container but the feature was not built. I noticed there was some skeleton code for this button container, so I removed it as well.


margaree (approved):
LGTM!

PR #951, opened 2022-07-12
Search: Enable Properties Part 4: Enable showing properties/events/slots in search results

margaree:
This is the final PR for enabling showing properties/events/slots in the Daylight search results.

Original PR https://github.com/BrightspaceUI/documentation/pull/918


tools/build-index-utils.js
```diff
@@ -10,34 +11,46 @@ export function buildRecordPreviewFromObject(url, pageTitle, sidenavSection, inf
...
-	if (!record || !sidenavSection || !type) {
+export function generatePreview(record, sidenavSection, preview) {
+	if (!record || !sidenavSection) {
 		throw new Error('generatePreview: A required parameter is missing from call');
 	}
+	preview = preview
+		? preview
+			.replace(/(\r\n|\n|\r)/gm, ' ')
+			.replace(/<(?!\/?code)[^>]*>/gm, '')
+			.replace(/&quot;/gm, '"')
```
margaree, line 32: `marked` is replacing quotes and & with `&quot;` etc which is making it through to the rendered text. Not sure if we have a suggested more universal approach for dealing with this

tools/build-index-utils.js
```diff
@@ -10,34 +11,46 @@ export function buildRecordPreviewFromObject(url, pageTitle, sidenavSection, inf
 	const id = `${url}-${urlSuffix}-${infoObject.name}`;
 
 	const content = `${infoObject.name} ${infoObject.type || ''} ${infoObject.description || ''}`;
+	let supportingName = infoObject.name || infoObject.attribute || (type === RECORD_TYPE.SLOT ? 'Default' : '');
+	if (supportingName) supportingName = `<code>${supportingName}</code>`;
+	const supportingDescription = `${infoObject.description && supportingName ? ': ' : ''}${infoObject.description || ''}`;
+	const previewText = marked(`${supportingName}${supportingDescription}`); // parse markdown into html
```
margaree, line 17: `marked` is used here to parse the markdown text in the description and name into html, which we can easily remove the tags and whitespace out of.

tools/build-index-utils.js
```diff
@@ -10,34 +11,48 @@ export function buildRecordPreviewFromObject(url, pageTitle, sidenavSection, inf
 	const id = `${url}-${urlSuffix}-${infoObject.name}`;
 
 	const content = `${infoObject.name} ${infoObject.type || ''} ${infoObject.description || ''}`;
+	let supportingName = infoObject.name || infoObject.attribute || (type === RECORD_TYPE.SLOT ? 'Default' : '');
+	if (supportingName) supportingName = `<code>${supportingName}</code>`;
+	const supportingDescription = `${infoObject.description && supportingName ? ': ' : ''}${infoObject.description || ''}`;
```
eKoopmans, line 16: I notice these lines 14-16 are essentially repeated in `tools/build-index.js`, wondering if there's any consolidation that could happen here?

eKoopmans (approved):

pages/_components/search-result.js
```diff
@@ -25,6 +25,7 @@ function applySearch(text, searchTerm) {
 	const splitString = text.split(re);
 
 	return splitString.reduce((accumulator, partialString) => {
+		partialString = partialString.replace(/<code>/g, '').replace(/<\/code>/g, ''); // remove unmatched code block tags
```
eKoopmans, line 28: Interesting, are you seeing this come up? I thought the regex would have caught any of these.

pages/_components/search-result.js
```diff
@@ -7,11 +7,12 @@ import { menuItemStyles } from '@brightspace-ui/core/components/menu/menu-item-s
 
 export function applyCodeAndSearch(text, searchTerm) {
 	if (!text || typeof(text) === 'object') return text;
-	const re = new RegExp('(<code>.*?</code>)', 'gi');
+	const re = new RegExp('(<code>.*?</code>)|(<code>.*?$)|(^(?:(?!<code>).)*?</code>)', 'gi');
```
eKoopmans, line 10: 😂 I love how complex `(?:(?!<code>).)*?` has to be. So essentially your addition matches any code blocks that are missing the trailing or leading tags? LGTM, though throwing a comment in might not go amiss 😄.

margaree: Right?! It definitely seems excessively complicated. I tried to simplify that piece but this was the simplest I got to that still worked. Yup, that's correct as to what it's doing.

PR #981, opened 2022-07-29
Fix broken internal content link

oelhanafey:
This link on the adding-component page was using an href from a preview site. Changed to use relative URL.


margaree (approved):
🤦‍♀️ good catch!

PR #988, opened 2022-08-05
Added dual focus to search and search results navigation

oelhanafey:
I wanted to gather feedback on my progress thus far on this. The search and search result navigation functionality mostly works with some minor bugs. I made some changes to the search components and will need to rewrite many of the tests so looking for feedback on this before I do that.

This PR changes the search components so that you can type in the search box while simultaneously navigating through the elements with the `UP` and `DOWN` arrow keys and navigating to the page with `ENTER`. This is done by keeping actual focus on the search-input component and storing the current search result focus position as a state variable.

Since we are keeping the actual focus on the search input, most of the menu navigation which was previously leveraging `<d2l-menu>` needs to be manually coded. This includes scrolling in new menu elements that are not currently in the view. This is done by the `<d2l-daylight-search-result>` component firing an event that is caught from the ancestor `<d2l-daylight-search>` component which uses the reference to the search result component to scroll it into view if needed.

To do:
- [x] Fix bug, navigating to other places within the same page with `#` does not scroll to the correct place
- [x] Fix bug, tabbing to daylight with search open will put focus on the dropdown content rather than the search input
- [x] Close menu when pressing `UP` on the first element, open menu when pressing `DOWN` (confirm with Jeff that this is still desired)
- [x] Add/change search component tests


pages/_components/search-result.js
```diff
@@ -184,6 +192,11 @@ class SearchResultMenuItem extends MenuItemMixin(LitElement) {
 				this._handleItemUpdated();
 			} else if (prop === 'searchTerm' && this.searchTerm && this.item) {
 				this._handleItemUpdated();
+			} else if (prop === 'focused' && this.focused) {
+				this.dispatchEvent(new CustomEvent('d2l-daylight-search-focus-change', {
+					bubbles: true,
+					composed: true
+				}));
```
oelhanafey, line 199: When a new `search-result` is focused, the component will fire a `focus-change` event so that the ancestor can catch it and get a reference to the focused element through `composedPath()`, scrolling it into view if needed

pages/_components/search.js
```diff
@@ -156,31 +181,45 @@ export class DaylightSearch extends LitElement {
 		});
 	}
 
+	_dropdownScrollToIfNeeded(element) {
+		const dropdownContent = this.renderRoot.querySelector('d2l-dropdown-content');
+		const container = dropdownContent.shadowRoot.querySelector('.d2l-dropdown-content-container');
```
oelhanafey, line 186: This function will scroll to the given element in the `dropdown-content` component. In order for this to work, it needs access to the scrolling container which is stored within the `dropdown-content` shadowDOM. This is not great, since it tightly couples this component with the inner workings of `dropdown-content` but I couldn't figure out another way to get the scrolling to work.

oelhanafey:
Couple things to clarify with Jeff:
1. He asked for pressing `UP` on the first element to close the menu. After testing this, I wasn't a huge fan of how it acted since holding up would navigate up and rather than stay on the first element it would close the menu. I would like to chat with Jeff about this once he's back but for the moment I did not add this. Pressing `DOWN` on a closed menu will open the menu though and you can still close the menu with `ESC`.
2. Closing and reopening the menu saves the current focus state. Should it reset to the top of the menu when re-opened?

I think these two changes are relatively minor and can wait until Jeff is back to be fixed in a separate PR if needed.

pages/_components/search.js
```diff
@@ -199,18 +237,31 @@ export class DaylightSearch extends LitElement {
 	}
 
 	_onKeyDown(e) {
-		if (!this._opened) return;
```
margaree, line 202: I noticed that when I reach the bottom of the results and press the down arrow key I no longer am brought to the top of the results. Was this on purpose?

margaree (commented):
I personally think it's fine that the focus state is saved between closing and re-opening, since it resets when the search changes, but Jeff would have a better idea on if that fits with existing design patterns/best practices.

pages/_components/search.js
oelhanafey: I did that on purpose but it was one of the things I was going to discuss with Jeff. I find that sometimes I want to just hold the down arrow to scroll to the bottom of the list and the wrapping prevents you from doing that. But I can add that behaviour back if we prefer the wrapping.

pages/_components/search.js
```diff
@@ -129,15 +155,13 @@ export class DaylightSearch extends LitElement {
...
 					@d2l-dropdown-close="${this._handleDropdownClose}"
 					@d2l-dropdown-open="${this._handleDropdownOpen}"
-					@focus="${this._handleDropdownContentFocus}"
 					min-width="800"
 					no-auto-focus
 					no-padding
 					no-padding-footer
 					no-pointer
-					?opened="${this._opened}"
-					trap-focus>
```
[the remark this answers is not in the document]
oelhanafey: Yeah I forgot to consider the tab ordering. I'll explore that and see if I can make that work more as you'd expect.

pages/_components/search.js
```diff
@@ -13,18 +13,22 @@ import { ifDefined } from 'lit-html/directives/if-defined.js';
...
 export class DaylightSearch extends LitElement {
 	static get properties() {
 		return {
+			_currentFocus: { state: true },
 			_hasResults: { type: Boolean, reflect: true, attribute: 'has-results' },
 			_opened: { type: Boolean },
 			_results: { type: String },
-			_searchTerm: { type: String }
+			_resultsLength: { state: true },
+			_searchTerm: { type: String },
```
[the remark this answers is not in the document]
oelhanafey: Maybe. There is a visual diff that calls `<d2l-daylight-search>` with the `_searchTerm` but that test could be altered to work. Tbh probably all of the properties should be states since they are all meant to be internal and the component is called with no attributes.

pages/_components/search-results.js
```diff
@@ -54,20 +55,31 @@ export class DaylightSearchResults extends LitElement {
...
+		if (!this._results || Object.keys(this._results).length === 0) {
+			return html`<div class="d2l-daylight-search-results-none">No results found</div>`;
+		}
+
+		let focusIndex = -1;
 
 		const numCategories = Object.keys(this._results).length;
 		const categories = Object.keys(this._results).map((category, index) => {
 			const menuItems = this._results[category].map((item) => {
+				focusIndex += 1;
```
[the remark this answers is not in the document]
oelhanafey: I thought about this for a bit but wasn't sure how to accomplish this while still using map since it is a 2D array. `map` doesn't increment the index like a for loop, it changes it to the appropriate array index, so I couldn't figure out a way to do this without an additional variable.

For example:
```
const test = [1,2,3];
test.map((item, index) => {
    test.map((item2) => {
        index += 1
        console.log(index);
    })
});
// 1 2 3
// 2 3 4
// 3 4 5
```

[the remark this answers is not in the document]
margaree: Ah right I completely didn't consider the nested map situation🤦‍♀️ nevermind!

pages/_components/search.js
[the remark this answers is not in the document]
oelhanafey: I changed all of the properties except `_searchTerm` to use `{ state: true }`. Changing the test that uses `_searchTerm` is a bit more complex and replacing the current test with a complex test that does the same thing seems to not be very productive. These components are only used by us within Daylight and having the component be easily testable is beneficial.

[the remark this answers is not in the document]
dlockhart: A workaround could be to just expose a method `setSearchTerm(term)` that the test uses, which sets the state variable. Good either way though.

pages/_components/search.js
```diff
@@ -102,8 +119,18 @@ export class DaylightSearch extends LitElement {
 	}
 
 	render() {
+		let searchResults = null;
```
dlockhart, line 122: Recommend using `nothing` here for this and `emptyState`.

pages/_components/search.js
dlockhart: Did you try [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) on the actual element?

oelhanafey: Yeah so I did try that and gave up on that since it was scrolling the page in addition to the menu. I just did another search today and it seems like I can fix that issue with:
```
element.scrollIntoView({ behavior: 'smooth', block: 'nearest', inline: 'start' });
```

But the other issue is that the [category header](https://github.com/BrightspaceUI/documentation/blob/cff125bbe0dfbf7477c303783ba120bcd2e57d7f/pages/_components/search-results.js#L37) uses sticky positioning and it doesn't scroll into view properly when you are scrolling up. This is why I added the `categoryHeaderOffset` in `_dropdownScrollToIfNeeded`. 

Screen Shot 2022-08-17 at 10 55 33 AM

oelhanafey: After investigating this further, the only way I found to fix this would be to manually scroll up by the `categoryHeaderOffset` after `scrollIntoView` finishes but you would also need to find when that additional scrolling would be necessary. This seems like it would end up being more complex than the current solution.

I did add an additional method `_getCategoryHeaderOffset` which will calculate the `categoryHeaderOffset` according to `font-size` rather than it only being correct at `font-size: 20px`.

dlockhart: I think I've solved this in the past by having a fake placeholder element that always takes up the same height as the sticky element? Not sure if that's an option here.

oelhanafey: Hmm I'm not sure how that would work. I can add an element with the same height as the offset but in order for it to be always present at the top of the menu it would need to be using absolute positioning which will take it out of the flow and not displace the search results, essentially the same result as using sticky positioning.

The way that I could see it working is if each `d2l-daylight-search-result` element was wrapped in an invisible container that had the `categoryHeaderOffset` included but that seems messy.

dlockhart: Hmm yeah your'e right. Just trying to think of any way of avoiding reaching into the shadow DOM here... could we maybe add a method to `<d2l-dropdown-content>` that does the scrolling for us?

oelhanafey: We could do that! We could add a method to `<d2l-dropdown-content>` that returns the scrolling container.

oelhanafey: Oh, yeah we could probably also move that whole scrolling method into `dropdown-content` so all the scrolling logic is done from there

dlockhart: Yeah either of those sound good to me!

dlockhart (approved):
LGTM!

PR #1007, opened 2022-08-16
Migrate to Node 16

oelhanafey:
Migrating Node version to 16. There were two main issues with the migration. 

In the build step, Daylight checks for any missing dependencies, installs them and adds their versions to the `dependencyVersions` list. `npm i ${pkg} --json` used to return a list of all package names and versions that were installed or updated, but in npm 8, this command only returns the number of installed or updated packages.

To fix this, I removed the `npm i` json parsing logic and instead added another build step after installing the dependencies, `_checkInstalledDependencies` which checks those recently installed dependencies, ensures that they are installed correctly and adds their versions to the `dependencyVersions` list.

The other issue was `npm i` deleting `devDependencies` in production which I expand on below.


tools/generate-pages.js
```diff
@@ -271,26 +293,14 @@ function _handleError(message) {
...
 }
 
-function _installDependencies(dependencies) {
-	if (!dependencies) return Promise.resolve();
+function _installDependencies(dependencyList) {
+	if (!dependencyList || dependencyList.length === 0) return Promise.resolve();
+	const dependencies = dependencyList.join(' ');
 	return new Promise((resolve) => {
-		exec(`npm i ${dependencies} --no-save --json`, (err, stdout) => {
+		exec(`npm i ${dependencies} --no-save`, (err) => {
```
oelhanafey, line 300: CI runs the build with `NODE_ENV=prod`. When in a prod environment, this `npm i` will delete all `devDependencies`. This can be solved by either changing this command to `NODE_ENV=dev npm i` or moving the `devDependencies` to `dependencies`.

I opted to move the `devDependencies` to `dependencies` since it seemed cleaner.

tools/generate-pages.js
```diff
@@ -241,24 +263,24 @@ function _getMissingDependenciesList() {
 	}
 
 	let dependencyCount = 0;
-	let dependencyString = '';
+	const dependencyList = [];
```
oelhanafey, line 266: This was causing some extra complexity in `_checkInstalledDependencies`. I changed this to a dependencyList and just parse it into a string when needed to simplify.

package.json
```diff
@@ -60,6 +61,8 @@
     "eslint-plugin-sort-class-members": "^1",
     "gray-matter": "^4",
     "lit-analyzer": "^1",
+	"lit-element": "^3",
```
dlockhart, line 64: Nit: looks like a tab instead of spaces here

dlockhart (approved):

PR #1014, opened 2022-08-18
Changed component page slack channel to #daylight-site

oelhanafey:
Daylight was linking to a private Daylight slack channel on the component pages. Changed to the public #daylight-site channel


dlockhart (approved):
🎉

PR #1048, opened 2022-09-06
fix: Add vertical margin to "best practices" list

jameswklassen:
## Description
The "best practices" section for Description List needs text before the actual best practices list.

This caused a spacing issue, since the `d2l-daylight-best-practices` has no vertical margin (like `<p>` tags do).

Adding this margin does not have a visual effect on cases where there's no text above the list.

image

### Before (no margin)
image

### After
image


oelhanafey (approved):
Oh, the reason I was hesitant about the other change was because I thought this would increase the margin for all the cases. 

Looks good!

eKoopmans (approved):

PR #1087, opened 2022-10-14
switch actions to use new output format

dlockhart:


.github/actions/update-labeled-issues-list/action.yml
```diff
@@ -16,7 +16,7 @@ runs:
         id: check-out-branch
         run: |
           BRANCH_NAME="ghworkflow/update-issue-list"
-          echo "::set-output name=branch-name::$(echo "$BRANCH_NAME")"
+          echo "branch-name=$(echo "$BRANCH_NAME")" >> $GITHUB_OUTPUT
```
eKoopmans, line 19: I've been wondering about these - is there a reason to re-echo, instead of just this?

```
echo "branch-name=$BRANCH_NAME" >> $GITHUB_OUTPUT
```

svanherk: I also cannot remember if this was on purpose or not 😛.  I removed a true/false double echo from the action I recently wrote based on somewhere that had done it, but I was set up to test it easily.

dlockhart: Ok I've removed them... 🤞

eKoopmans (approved):

.github/actions/update-labeled-issues-list/action.yml
svanherk: I _wonder_ if it affects this line on whether or not it needs quotes: https://github.com/BrightspaceUI/documentation/pull/1087/files#diff-5671ef4b7c45837c1c51cb06549539cadb28319591c25ac63f9f00b56a799ff8R60

eKoopmans: @svanherk I'm not sure which line that's linking to, it's not showing up for me - but I'm pretty sure we're good.

I had a bit of a chat with Dave about quotes, they can be finicky in Bash, but shouldn't have an impact here. Double quotes don't do a whole lot for echo, everything's treated as a string regardless - here's a bit of a discussion:
https://unix.stackexchange.com/a/269232

Single quotes would be different, they force things to be treated literally, which we wouldn't want.

svanherk: Not sure why GitHub lets me copy a link to a collapsed line if its not going to work to actually link there 😒 But it was this line:
```
if [ ${{ steps.committing-changes.outputs.changes }} == false ]; then
```
But seems like we should be good, we're not really changing anything

eKoopmans: ⚠️ Okay, I have identified a potential risk around quotes, though it's not really related to `$()`. The good news is it's only relevant for values with whitespace, which most our values don't, but some might (`$FAILED_REPORTS`?). I've already looped in @dlockhart, I think we're on the same page.

Here's what it looks like:

```bash
bar=string with whitespace   # Fails: "with: command not found"
bar="string with whitespace" # Passes
echo "foo=$bar" # Echoes "foo=string with whitespace", which would fail if executed
echo "foo=$(echo "$bar")" # Same thing
echo "foo=\"$bar\"" # Echoes "foo="string with whitespace"", which should pass
```

So, for any strings we suspect might have whitespace, we should probably be throwing safety `\"` around them. It's weird that this isn't called out on [the docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter), but maybe they've done some magic to prevent it being an issue?

---

One other tiny side-note: using `$()` *could* have an effect in this situation:

```bash
bar="weird    whitespace"
echo "foo=$bar" # foo=weird    whitespace
echo "foo=$(echo "$bar)" # Same
echo "foo=$(echo $bar)" # foo=weird whitespace
```

The last example "normalizes" the whitespace - but I don't think we're deliberately doing that anywhere, and regardless we'd need to add `\"` around any strings with whitespace anyway.

PR #1122, opened 2022-11-15
Add Daylight Code Sandbox

oelhanafey:
[US145476](https://rally1.rallydev.com/#/?detail=/userstory/670705972601&fdp=true)

This PR adds the Code Sandbox to Daylight so that users can test out Daylight web components in the browser. This was introduced using Google's [playground-elements](https://github.com/google/playground-elements) library that we also use for our demos on component pages so no additional dependencies were necessary. 

I used the all-in-one component `<playground-ide>` which contains a code editor, file editor and demo preview in a single web component which made it easy to implement but also provided limited styling options. In the future, it would be nice to break this into its individual components to customize further such as changing the `flex-direction` to column on small viewports.


pages/_includes/layouts/base.njk
```diff
@@ -1,10 +1,16 @@
...
 	{% include "css/design-tabs.css" %}
 	{% include "css/nav.css" %}
 	{% include "css/page-content.css" %}
 	{% include "css/style.css" %}
+
+	{% if sandbox %}
+		{% include "css/sandbox.css" %}
+	{% endif %}	
 {% endset %}
+{% set hasFooter = eleventyNavigation.key != 'sandbox' %}
```
oelhanafey, line 13: I explored creating a separate layout for the sandbox since it does not use a footer but there ended up being a lot of repetition. I opted to add the imports and css to `base.njk` conditional that it is loading the sandbox page, so it won't import these files if they aren't needed.

I decided to create a `hasFooter` variable so that if someone wanted to create other pages with no footer, they could just add them here.

pages/_components/code-sandbox.js
```diff
@@ -0,0 +1,115 @@
...
+const url = window.location;
+
+// The sandboxUrl acts as a host for our demo to load component files (generated in this repo as playground-build)
+const sandboxUrl = url.host.includes('localhost') ? 'http://localhost:8081' : `${url.protocol}//playground-${url.host}`;
+
+const indexHTML = `
+<!DOCTYPE html>
+<html>
+  <head>
+    <!-- Import the stylesheet and web components -->
```
eKoopmans, line 15: When using this for the first time, I wasn't sure if I had to add my own imports, something like this might help:

```suggestion
    <!-- Import the stylesheet and web components - no addition required -->
```

pages/_components/code-sandbox.js
```diff
@@ -0,0 +1,115 @@
...
+	static get styles() {
+
+		return css`
+			#playground {
+				border-bottom: none;
+				box-sizing: border-box;
+				height: 100vh;
+			}
+
+			${unsafeCSS(generateOverlayVariables())}
```
eKoopmans, line 85: Does this need to be unsafe CSS? Could `generateOverlayVariables` return ``css`...` ``?

eKoopmans (commented):
LGTM! Just two comments. I also noticed some issues on Firefox, don't need to resolve them now but something to keep an eye on:

1. Demos load much slower in FF than Chrome (~2+ seconds for each change, as opposed to almost instant)
2. In FF, the cursor sometimes disappears in the code editor - e.g. after clicking the "refresh" icon in the top right, then clicking into the code editor

pages/_components/code-sandbox.js
```diff
@@ -0,0 +1,115 @@
+import 'playground-elements/playground-ide';
```
dlockhart, line 1: Should this have a `.js` extension? Likely works with `web-dev-server`, but may not in-browser.

pages/_components/code-sandbox.js
dlockhart: I'd be interested in exploring a solution where either you _did_ have to specify your own imports, or we could detect them automatically.

Specifically worried about:
- This imports a ton of stuff that won't be needed in most situations, hurting performance
- As we add new things to `@brightspace-ui/core`, we're going to need to remember to add them here. Ultimately we're going to forget, and someone will try to use something that isn't here.
- What about non-core things like labs components?

pages/_components/code-sandbox.js
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	render() {
+
+		return html`
+			<playground-ide id="playground" editable-file-system line-numbers resizable sandbox-base-url="${sandboxUrl}">
+				<script type="sample/importmap">
+					{
+						"imports": ${JSON.stringify(IMPORTS)}
+					}
```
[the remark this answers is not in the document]
dlockhart: Neat!

pages/_includes/layouts/base.njk
dlockhart: Wondering... could `hasFooter: false` be part of the layout's front-matter, and then `sandbox.njk` could override that to `true`? If that works, we could do something similar for `cssIncludes: []` and `jsIncludes: []` as a mechanism for child layouts or pages to adjust the template.

pages/_includes/layouts/sandbox.njk
```diff
@@ -0,0 +1,5 @@
+---
+layout: layouts/base
+permalink: "sandbox/"
+---
+<d2l-daylight-code-sandbox></d2l-daylight-code-sandbox>
```
dlockhart, line 5: Is this layout even needed? i.e. could this be merged into `sandbox/index.md`?

pages/_components/code-sandbox.js
oelhanafey: I hope it works because that's how all the imports are done for the component demos 😬 

I'll fix that, thanks!

pages/_components/code-sandbox.js
oelhanafey: 1. My concern was making this as easy to use as possible. My thought was the performance hit was worth the convenience.
2. Core components will automatically be added here. The files are pulled from the `manifest-wc.js` file that rollup creates so no manual intervention required, provided that the issue on GitHub was created for the component as is already required.
3. You can't import labs components at the moment on Daylight. It was never setup to install those repos. My guess was that it was too much work because they were all stored in separate repos. This might be possible to implement now with the `BrightspaceUI/labs` repo.

That was my thought process for it but I could certainly change it if we want to maximize performance. If we do update Daylight to include labs components too, then it would end up importing even more.

dlockhart: Ok thanks for the explanation. I'm good with this as it is, and we can adjust later if performance is an issue.

pages/_components/code-sandbox.js
oelhanafey: Yep I'll see if I can avoid using the `unsafeCSS`. Thanks!

pages/_includes/layouts/sandbox.njk
oelhanafey: Yep this worked. The only issue was that the markdown added a `<p>` tag around `<d2l-daylight-code-sandbox>` automatically which had some margins from `.d2l-typography`. Adding a wrapping `<div>` fixed this.

pages/_includes/layouts/base.njk
oelhanafey: Yep this works! Though instead of adding a generic `jsIncludes` I needed to create one specifically for components, `componentImports`, because it needs to use JS to get the full pathname which doesn't work from the front-matter. If another kind of import needs to be added in the future it can be added in a similar way.

pages/_components/code-sandbox.js
oelhanafey: `unsafeCSS` is needed here because I am not using literals and adding the colors using a variable. I could change it to use literals but I like it this way since if we wanted to add a light mode, it would be easy to extend from this. This should be safe as it is only using colours pulled from our own core repo.

pages/_includes/layouts/base.njk
```diff
@@ -1,9 +1,16 @@
+{% if hasFooter == null %}
+	{% set hasFooter = true %}
+{% endif %}
```
dlockhart, line 3: Could also invert this to `hideFooter` if we'd prefer it be false-by-default. 🤷

dlockhart (approved):
Looks awesome!

PR #1124, opened 2022-11-15
use gzip compression for certain assets

dlockhart:
Attempting to enable GZIP compression for HTML, CSS and JS assets.

We'll see if this works...


.github/workflows/ci.yml
```diff
@@ -42,7 +42,10 @@ jobs:
           role-duration-seconds: 3600
           aws-region: ca-central-1
       - name: Deploy - branch
-        run: aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}"
+        run: |
+          find ./_site/ \( -iname '*.html' -o -iname '*.css' -o -iname '*.js' -o -iname '*.json' -o -iname '*.svg' \) -exec gzip -9 -n {} \; -exec mv {}.gz {} \;
+          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'gzip' --exclude '*.*' --include '*.html' --include '*.css' --include '*.js' --include '*.json' --include '*.svg'
+          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --exclude '*.html' --exclude '*.css' --exclude '*.js' --exclude '*.json' --exclude '*.svg'
```
dlockhart, line 48: @oelhanafey: I guess I should do the playground bucket too, eh?

oelhanafey: Yeah, that would probably be a good idea too.

It looks like you are only gzipping certain file types. What is the reason for the exceptions?

dlockhart: Only text-based filetypes that don't already have some kind of compression algorithm built into them need to be compressed. So things like PNG/JPEG you don't compress. If you can think of other extensions the site uses that I'm missing, let me know!

oelhanafey: Ah okay thanks! Then yeah, the playground bucket should be all `.js` files, so it would be good to gzip them as well.

oelhanafey (approved):
Nice!

.github/workflows/ci.yml
dlockhart: Ok with the playground compressed, I re-ran some numbers on the "button component" page instead of the home page. That page has 86 requests and 17 MB of resources!

Before unprimed: 17 MB transferred, 2600ms finish time, 2500ms load time
After unprimed: 4.1 MB transferred (76% reduction), 1400ms finish time (46% quicker), 1300ms load time (48% quicker)

PR #1125, opened 2022-11-15
use cache-control headers for JS assets

dlockhart:
Continuing on from #1124 which enabled `gzip` compression, this sets the `cache-control` header for our JavaScript assets.

The reason we can do this is that the build hashes the contents of our JS files into their filenames, so whenever they change the filename changes which will invalidate any browser caches.

In testing, this results in 24 fewer requests on the homepage and 62 fewer requests on the "button" component page.


oelhanafey (commented):
There appear to be 2 exceptions to the hashing. On the main site, the `lunr.js` is not hashed and on the playground, the `playground-service-worker.js` is not hashed so updates to these libraries might not be downloaded if the user has the site cached.

oelhanafey (approved):

dlockhart:
Yeah CloudFront may be overkill here since we don't really care about geo-replication (for now). Sweet!

PR #1128, opened 2022-11-16
Fix Daylight page height

oelhanafey:
[DE50926](https://rally1.rallydev.com/#/?detail=/defect/671356987941&fdp=true)

The Daylight site currently has a scrollbar appear even when all of the page contents can fit on the viewport. This is due to the height of some components not being calculated correctly. You can see it [here](https://daylight.d2l.dev/about/overview/) if you maximize your screen.

This PR fixes the height of the page components so that there are no scrollbars when the contents can fit on the screen. The appearance of the page shouldn't significantly change but there may be some small pixel differences.


eKoopmans (commented):
So... there's a lot of numbers changing here and I'm not sure why 😅. Are all the changes required? When I open the preview and try disabling some of the styles (like the heights on `title-image`, `search-container`, etc) it doesn't *appear* to be breaking anything, but I'm sure I'm missing something.

I imagine lots of tweaking was required, and I trust that the changes are all for good reasons. I'm tempted to say "sure looks good", but I could use more context on why each of the changes are happening.

A few things that stand out to me:

- what was actually causing the underlying sizing issue you mentioned?
- why `- 2px` in the side panel? (and corresponding 2px changes in `nav` and `main-content`)
- why the explicit heights?
- why `line-height: 0` on title and footer?
- why the mix of px and rem in some calcs?
- why adding 0.35rem padding (bottom) in title, and 0.3rem (top and bottom) in footer?

oelhanafey:
So there were two separate things causing the underlying sizing issue. First issue was with the sidebar height being more than `100vh` and the other was because the footer was bigger than the space allotted to it. There was a lot of tinkering but I'll break down the changes that I made.

1. The `-2px` in the side panel is due to it having `1px` top bottom borders which were not accounted for.
2. I don't think the explicit heights are mandatory. I kept them at the size that they were before my changes but this makes it a bit more resilient. For example, if the SVG gets changed and that SVG is a different size, it could mess with the rest of the calculations so I just added them explicitly.
3. The `img` tags are inline by default and are subject to having a `line-height`. The result of that was, for example the 16px img having some spacing above and below it. This made it very difficult to accurately make calculations on the height of the component. Making the `line-height: 0` makes it so that the `` tag dimensions are exactly the size of the image itself.
4. The mix of px and rem calculations are due to the size of the images and search container being a static pixel height but the margins were set using rem. This caused changes in `font-size` to mess up the height calculations. The updated calculations should work now given any `font-size`.
5. The addition of some extra padding to the footer and title were due to the aforementioned `line-height` issue. It caused some extra space around the footer and below the title. I wanted to keep the appearance of the page as close as possible to what it was before, so I added those in place.

eKoopmans:
Awesome, I appreciate the explanations @oelhanafey.

> The -2px in the side panel is due to it having 1px top bottom borders which were not accounted for.

Did you try using `box-sizing: border-box` instead? Usually cleaner than having to calculate around the borders.

oelhanafey:
@eKoopmans Thanks, yeah that is cleaner! I changed it to use `border-box` and also changed the added footer padding to be placed around the img rather than the footer. There is an optional div at the bottom of the footer detailing the version of core that was used for that version of Daylight. I don't think we actually use it anymore but this should prevent any regression in the visual-diffs from the `line-height: 0` change.

pages/_includes/css/style.css
```diff
@@ -43,7 +43,7 @@ div.d2l-daylight-content {
 /** Main Panel */
 .d2l-daylight-main-content {
 	grid-area: content;
-	min-height: calc(100vh - 24px - 3rem - 4.5rem); /* 100vh - min height of footer including padding and margin */
+	min-height: calc(100vh - 16px - 3.6rem - 4.5rem); /* 100vh - min height of footer including padding and margin */
```
eKoopmans, line 46:
```suggestion
	min-height: calc(100vh - 16px - 3.6rem - 4.5rem); /* 100vh - min height of footer including padding and grid row-gap */
```

pages/_includes/css/nav.css
```diff
@@ -34,7 +41,7 @@ nav .d2l-link:not(.d2l-daylight-link-selected) {
 /** Side nav */
 nav {
 	display: block;
-	height: calc(100vh - 204px); /* 100vh - height of title + search */
+	height: calc(100vh - 118px - 4.4rem); /* 100vh - height of title + search */
```
eKoopmans, line 44: 4.4rem checks out, but why 118px? I'm getting 115 from `title-img` and `search-container`. Is the other 3 from borders again (1px on `search-container` bottom, 2px from ??).

oelhanafey: Yeah it is. 1px from `search-container` and 2px from `side-panel`. It is subtracting from `100vh` so I'm not aware of any way that I can avoid measuring around the borders in this case without overhauling a lot of the sidebar CSS. I added it to the comments.

eKoopmans (approved):
Cool, thanks for the explanations Oscar!

PR #401, opened 2021-11-02
New favicons + About Daylight section changes

geurts:
Adds an SVG favicon and a title to the base.njk file, and adds some new About Daylight pages (Margarees 'how to contribute' docs and some stubs) along with the fancy Designing / Developing links to the Overview page.

PR #1133, opened 2022-11-17
fix: put page name before site name in title

KaiPrince:
[DE50946](https://rally1.rallydev.com/#/?detail=/defect/671452384475&fdp=true): Daylight site should have page name in front of site name

It is considered accessibility best practice to put the page-specific name in front of the site name. The advantage being that a list of open tabs wouldn't all start with "Daylight".


oelhanafey (approved):
Screen Shot 2022-11-17 at 2 35 26 PM

Thanks! This does appear to be the standard. This will be helpful for me too 😆

pages/_includes/layouts/base.njk
```diff
@@ -10,7 +10,7 @@
 <head>
 	<meta charset="utf-8">
 	<meta name="viewport" content="width=device-width, initial-scale=1.0">
-	<title>Daylight -&nbsp;{%- if eleventyNavigation.title -%}{{ eleventyNavigation.title }}{% else %}{{ title }}{%- endif %}</title>
+	<title>{%- if eleventyNavigation.title -%}{{ eleventyNavigation.title }}{% else %}{{ title }}{%- endif %} -&nbsp;Daylight</title>
```
eKoopmans, line 13: Not sure the specific reason for one ` ` and one `&nbsp;`, any particular reason to put them in the order you have? It seems more consistent with the original to use `&nbsp;- Daylight`.

oelhanafey: My thinking was that the way @KaiPrince did it was more consistent with the original, since it was probably added so that the dash and second title don't wrap separately 🤔 

I'm not sure the reasoning for it. It was added as part of [this PR](https://github.com/BrightspaceUI/documentation/pull/401) but no reason was provided.

PR #1137, opened 2022-11-24
allow caching of image assets

dlockhart:
This allows us to set "cache forever" headers on all the Daylight Site images. It accomplishes this by passing a version on the query-string (`?v=1` for the first version), which we'll need to bump if any of these images ever change in the future. That's slightly annoying, but IMO it's worth it.

Once I move the search to OpenSearch, these images are pretty much the only remaining cacheable things that aren't currently being cached.

While I was in here, I removed the local `favicon` copies and just referenced them from the CDN.


.github/workflows/ci.yml
```diff
@@ -43,19 +43,19 @@ jobs:
...
-          aws s3 sync --delete ./playground-build/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}-playground-build" --content-encoding 'br' --exclude '*.*' --include '*.html' --include '*.css' --include '*.json' --include '*.svg' --include 'playground-service-worker.js'
+          find ./playground-build/ \( -iname '*.html' -o -iname '*.js' \) -exec brotli {} \; -exec mv {}.br {} \;
+          aws s3 sync --delete ./playground-build/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}-playground-build" --content-encoding 'br' --exclude '*.*' --include '*.html' --include 'playground-service-worker.js'
           aws s3 sync --delete ./playground-build/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}-playground-build" --content-encoding 'br' --cache-control "public,max-age=31536000,immutable" --exclude '*.*' --include '*.js' --exclude 'playground-service-worker.js'
-          aws s3 sync --delete ./playground-build/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}-playground-build" --exclude '*.html' --exclude '*.css' --exclude '*.js' --exclude '*.json' --exclude '*.svg'
```
dlockhart, line 58: The playground only has HTML and JS files, so simplifying this logic.

.github/workflows/ci.yml
```diff
@@ -43,19 +43,19 @@ jobs:
           aws-region: ca-central-1
       - name: Deploy - branch
         run: |
-          find ./_site/ \( -iname '*.html' -o -iname '*.css' -o -iname '*.js' -o -iname '*.json' -o -iname '*.svg' \) -exec brotli {} \; -exec mv {}.br {} \;
-          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'br' --exclude '*.*' --include '*.html' --include '*.css' --include '*.json' --include '*.svg'
-          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'br' --cache-control "public,max-age=31536000,immutable" --exclude '*.*' --include '*.js'
-          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --exclude '*.html' --exclude '*.css' --exclude '*.js' --exclude '*.json' --exclude '*.svg'
+          find ./_site/ \( -iname '*.html' -o -iname '*.js' -o -iname '*.json' -o -iname '*.svg' \) -exec brotli {} \; -exec mv {}.br {} \;
```
dlockhart, line 46: There's no CSS files, so removing that.

pages/_components/component-owner.js
```diff
@@ -41,7 +41,7 @@ export class DaylightComponentOwner extends LitElement {
 		if (!this.owner || this.owner === 'Gaudi' || this.owner.includes('#')) {
 			const slack = (!this.owner || this.owner === 'Gaudi') ? GAUDI_SLACK : this.owner;
 
-			const icon = this.showSlackIcon ? html`<img src="/img/slack-logo.svg" alt="" />` : null;
+			const icon = this.showSlackIcon ? html`<img src="/img/slack-logo.svg?v=1" alt="" />` : null;
```
eKoopmans, line 44: Oof, I don't know about this cache-busting built straight into the code. I'm not personally a fan. A few thoughts:

- it feels like caching behaviour shouldn't be a concern of the source code
- it may not be obvious to people reading/contributing to the code, what the purpose of the query param is
- risk of forgetting or not knowing to bump the version when changing images (and/or not adding a version when adding new images)

Is there an alternative, maybe a rollup module that will rename all the image files by adding a hash to their name?

dlockhart: Yeah I don't love it either... I'll have a look. [rollup-plugin-static-files](https://www.npmjs.com/package/rollup-plugin-static-files) looks promising...

dlockhart: Negative, only works for `<script>` and `<link>`.

dlockhart: Ok I'm giving up on this for now due to time constraints, but I've reverted everything except the favicon change and removing CSS files from the s3 publish (Daylight has no CSS files).

eKoopmans (approved):
Sorry for being a downer about image caching!

dlockhart:
> Sorry for being a downer about image caching!

Haha all good. Deep down I knew it was a bad idea, just needed someone to point it out.

PR #1153, opened 2022-12-12
Add summary and more FAQs to readme

eKoopmans:
Adding two pieces to the readme:

1. A summary - recently on #web-platform we had someone searching this repo for documentation, without understanding that this was just the codebase and the actual docs were on [Daylight](https://daylight.d2l.dev/about/overview/). Hopefully this will clarify
2. Some FAQs related to my debugging of the bugged selection page after adding `selection-header`


README.md
```diff
@@ -106,3 +108,27 @@ The tooling that assembles the Daylight site makes several GitHub API calls. In
...
+
+Note that these changes are built and will *not* live-update - you will have to stop and restart with `npm start` if you want to test additional changes.
+
+As an alternative, you can create a new markdown file in `pages/` and copy-paste your readme into it, which will then be subject to live updates. If doing this, remember to add an appropriate metadata block to the start of the file (see any other `.md` file in `pages/` as an example).
+
+**Can I run the Daylight site locally without the step of reading all GitHub issues?**
+
+Yes, partially. You can build and run just the local markdown (e.g. everything in `pages/`) by replacing the contents of `_requestIssues()` in `generate-pages.js` with `return [];`.
+
+However, this strategy is limited, and notably you *cannot* use any web components from core etc this way unless making additional configuration changes. This is because the web component imports for use in Daylight are collectively pulled from the metadata of all GitHub issues.
```
oelhanafey, line 134: I'm not sure we should call this out in the README. I think it might be better if we just tell everyone to setup the GItHub token as part of the setup, at least for now as we might change this in the future.

README.md
```diff
@@ -106,3 +108,27 @@ The tooling that assembles the Daylight site makes several GitHub API calls. In
...
+
+This is not very convenient for debugging and testing changes. Instead, you can:
+
+1. Install dependencies as normal: `npm i`
+1. Find your component's readme file (or other files to change) in `node_modules`
+1. Make any changes, then run with `npm start` to see the changes
+
+Note that these changes are built and will *not* live-update - you will have to stop and restart with `npm start` if you want to test additional changes.
+
+As an alternative, you can create a new markdown file in `pages/` and copy-paste your readme into it, which will then be subject to live updates. If doing this, remember to add an appropriate metadata block to the start of the file (see any other `.md` file in `pages/` as an example).
```
oelhanafey, line 128: This is useful! Maybe also add that if you want to test a newly created component that is not currently being imported into the Daylight site, you can add it to the `componentFiles` list in `rollup.config.js`. Alternatively, that information could also be added to the "Why isn't my component being imported correctly?" information.

README.md
eKoopmans: Updated to say the GH build step is required.

README.md
eKoopmans: Done, added to "Why isn't my component being imported correctly".

README.md
```diff
@@ -106,3 +108,27 @@ The tooling that assembles the Daylight site makes several GitHub API calls. In
 1. Generate a new GitHub token (Settings > Developer Settings > Personal access tokens). It does not need any permissions. Copy this token.
 1. Create a `.env` file in your local repo.
 1. In your `.env` file add `GITHUB_TOKEN=<your new token>`
+
+**Why isn't my component being imported correctly?**
+
+If your component is failing to load on its corresponding readme, double-check that you have added it to the `components` array in your component's [issue on BrightspaceUI/documentation](https://github.com/BrightspaceUI/documentation/issues).
```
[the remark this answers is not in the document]
oelhanafey: I think the `Contributing` section of the site should contain all information about adding and editing components since it already contains most of the detailed information on the topic and we don't want people to have to go to two places to get all the information. But it looks like it already has some information about [adding new components](https://daylight.d2l.dev/about/editing-page/#adding-a-new-component-to-an-existing-page). If it makes it easier to find the information, then I think it's okay to have it duplicated here.

The other two are related to getting the repo up and running and testing changes so I think they make sense directly in the component README.

oelhanafey (approved):
Thanks!

PR #1162, opened 2022-12-19
Set token for octokit calls explicitly

svanherk:
Looks like the token was not being explicitly set and we were only getting a rate limit of 60 calls an hour before.  Seems like maybe this was shared between all unauthenticated calls against a runner.  This PR increases the page size so we need to make less calls, and passes the token in so our rate is raised to `15000`.


tools/generate-pages.js
```diff
@@ -338,10 +338,14 @@ function _processIssues(issues) {
 }
 
 async function _requestIssues() {
+	const rateLimit = await octokit.request('GET /rate_limit', {});
+	console.log(`Rate limit total: ${rateLimit.data.rate.limit}`);
+	console.log(`Rate limit remaining: ${rateLimit.data.rate.remaining}`);
```
svanherk, line 343: I figured it'd be nice to leave this for a while in case we need to investigate, but I could also take it out.  It doesn't count against your actual rate limit

oelhanafey:
Should we add the token to the [visual-diff workflow](https://github.com/BrightspaceUI/documentation/blob/main/.github/workflows/visual-diff.yml) as well? It also runs a build.

oelhanafey (approved):
Thanks Stacey!

svanherk:
> Should we add the token to the [visual-diff workflow](https://github.com/BrightspaceUI/documentation/blob/main/.github/workflows/visual-diff.yml) as well? It also runs a build.

Thanks missed that!  Perfect example of what it looks like without passing the token:
image

eKoopmans (approved):
Thanks!

PR #1175, opened 2023-01-04
Fix: Apply interactive attributes inside scripts for live demos

eKoopmans:
Daylight's "live" demos currently have a defect when targeting an element that is nested in a shadow DOM (e.g. `name:d2l-table-wrapper` when it's inside a wrapper component):

✅ It displays the correct properties, and editing them updates them in the code snippet
❌ Editing them does *not* update the actual live demo

I made three fixes in this PR:

1. The wrapper components are defined in `<script>` blocks, which are handled separately. These scripts were not getting the same attribute updates as the rest of the "parsed code"
2. Attribute defaults are read from the code snippet. The way they're determined wasn't working for elements within a shadow DOM
3. Also made a small fix to not lint `.gitignore` files

Rally: https://rally1.rallydev.com/#/?detail=/userstory/675086171663&fdp=true


pages/_components/demo-resizable-preview.js
```diff
@@ -460,12 +460,14 @@ class DaylightDemoResizablePreview extends FocusVisiblePolyfillMixin(LitElement)
...
-		if ((changedProperties.has('code') || (changedProperties.has('_visible') && this._visible)) && this._preview && this._project) {
+		if ((changedProperties.has('code') || changedProperties.has('scripts') || (changedProperties.has('_visible') && this._visible)) && this._preview && this._project) {
 			// Updates to the project must be directly applied via JS as playground elements
 			// do not update when slots are changed. Firing saveDebounced forces the playground to reload
 			// once the file content has been updated
 			const indexFile = this._project.files.find(({ name }) => name === PREVIEW_FILE_NAME);
 			indexFile.content = this.indexHTML;
+			const scriptsFile = this._project.files.find(({ name }) => name === 'index.js');
+			scriptsFile.content = this.scripts;
 			this._project.saveDebounced();
```
eKoopmans, line 471: Not sure if this is the most efficient approach here, but it works. I could probably split up the updates to `indexFile` and `scriptsFile` separately if we're concerned about performance.

eKoopmans:
P.S. This is the fix @dlockhart was proposing - with it, we should be able to have live demos target a nested (shadow DOM) component and modify it directly, without needing any additional passthrough wrapper.

pages/_components/demo-snippet.js
```diff
@@ -133,7 +133,7 @@ class DaylightDemoSnippetWrapper extends LitElement {
 	}
 
 	get scripts() {
-		const content = /<script.*>([\s\S]*)<\/script>/g.exec(this._demoSnippet);
+		const content = /<script.*>([\s\S]*)<\/script>/g.exec(this.code);
```
oelhanafey, line 136: What does this do?

pages/_components/demo-resizable-preview.js
```diff
@@ -460,12 +460,14 @@ class DaylightDemoResizablePreview extends FocusVisiblePolyfillMixin(LitElement)
...
 	}
 	update(changedProperties) {
-		if ((changedProperties.has('code') || (changedProperties.has('_visible') && this._visible)) && this._preview && this._project) {
+		if ((changedProperties.has('code') || changedProperties.has('scripts') || (changedProperties.has('_visible') && this._visible)) && this._preview && this._project) {
 			// Updates to the project must be directly applied via JS as playground elements
 			// do not update when slots are changed. Firing saveDebounced forces the playground to reload
 			// once the file content has been updated
 			const indexFile = this._project.files.find(({ name }) => name === PREVIEW_FILE_NAME);
 			indexFile.content = this.indexHTML;
+			const scriptsFile = this._project.files.find(({ name }) => name === 'index.js');
```
oelhanafey, line 469: Would it be useful to have a `SCRIPT_FILE_NAME` constant like we have for `PREVIEW_FILE_NAME`?

pages/_helpers/utils.js
```diff
@@ -40,8 +40,13 @@ export function getCode(codeSnippet, interactive, currAttributes, tagName) {
...
 	if (!code || !tagName) return {};
 
+	const re = new RegExp(`<${tagName}( [^>]*)?>`);
+	const match = code.match(re);
+	if (!match || !match[0]) return {};
+	const simplifiedComponent = `${match[0]}</${tagName}>`;
+
 	const t = document.createElement('template');
-	t.innerHTML = code;
+	t.innerHTML = simplifiedComponent;
```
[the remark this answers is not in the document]
dlockhart: Seems like it's the shadow root thing, which makes sense. Parsing HTML with a regex always [reminds me of this](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags) lol... 

This is a pretty limited case though so I'm sure it's fine, but alternately you could likely search for it programatically once you've put the whole code block into the template. That could get tricky waiting for the component(s) to render though.

pages/_components/demo-snippet.js
eKoopmans: So, regardless of the change this line is extracting the contents of the last `<script>` tag in the demo snippet, and storing it as `content`. Instead of looking at the raw snippet, I've changed it to look at `code`, which is:

- computed with [getCode(...)](https://github.com/BrightspaceUI/documentation/blob/main/pages/_helpers/utils.js#L9)
- identical to `_demoSnippet` if the snippet isn't interactive
- injects the interactive attributes from `this._attributes` into the first regex match of `this.tagName`

If you try the interactive demo for `d2l-table-header` in the current live version of Daylight, you'll see that the text in the code snippet updates when you use the interactive table - that's `this.code`. But the raw snippet without those attributes was previously being used to build the scripts to send into the actual interactive demo.

pages/_helpers/utils.js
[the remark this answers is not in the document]
eKoopmans: Yeah it's a shadow root thing. I tried mucking around with searching for it but the nodes I was getting were weird/different, I couldn't find a shadowRoot - either because it's a template or because it's not attached to the DOM, I didn't investigate too far.

Basically the same regex pattern is already being used in `getCode`, so it didn't seem like too much harm...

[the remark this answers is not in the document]
dlockhart: Gotcha, makes sense -- thanks.

dlockhart (approved):
Nice, thanks for fixing this up!

oelhanafey (approved):
Nice, thanks Erik!

test/components/component-utils.test.js
```diff
@@ -159,7 +159,23 @@ describe('component-utils', () => {
...
+						class Foo extends LitElement {
+							render() {
+								return html\`<d2l-button text="welcome here"></d2l-button>\`;
+							}
+						}
+					</script>
+					<d2l-nested-example></d2l-nested-example>
+				`,
+				tagName: 'd2l-button',
+				expected: { 'text': 'welcome here' }
```
eKoopmans, line 177: ??? Just realized this test shouldn't be passing (but is) - I never define `d2l-nested-example`! Also I meant to rename the class from `Foo` 🤦.

eKoopmans: Oh, right. Because `getDefaults` now doesn't actually render the whole snippet, it just extracts the `d2l-button` to grab its attributes. I'll still fix up the example though.

oelhanafey (approved):

PR #1187, opened 2023-01-10
Add null check for non-existent tags

oelhanafey:
Adding a README demo with a non-existent tag-name was causing Daylight to crash. This adds a null check to make sure that the demo has an entry in the `custom-elements-all.json` file. If it does not, then it will not display the properties, events and slots table for that demo but it will no longer crash the whole site.


dlockhart (approved):
👍

PR #1267, opened 2023-03-15
replace focus with focusWithKeyboard

dlockhart:


eKoopmans (approved):
Looks good! The diffs are because `resetFocus` was causing these focus tests to be mouse-based previously.

PR #1270, opened 2023-03-16
upgrade to Node 18

dlockhart:
The new lock-file version is backwards compatible with version 2, so this is safe to merge and won't impact anyone still on Node 16.


svanherk (approved):
Loving these `package-lock.json` changes 🎉🎉🎉

PR #1296, opened 2023-04-10
Fundamentals section with internationalization

dlockhart:
I played around with a few ideas on where to put an "internationalization" section and this is where I landed:

- Adds a new top-level "Fundamentals" area, of which Internationalization is a member
- Internationalization incorporates the existing bidirectionally bits from "style elements" and adds information for translations
- Moves "Accessibility for design" into "Fundamentals", dropping the "for design"

The idea here is that "fundamentals" span both dev and design. This is similar to [Shopify > Foundations](https://polaris.shopify.com/foundations). I didn't feel "How-tos" really had the right feel to it.

Open to other ideas & suggestions!


pages/fundamentals/internationalization.md
```diff
@@ -0,0 +1,57 @@
...
+  parent: fundamentals
+---
+
+# Internationalization
+
+The term "internationalization" covers everything involved in building interfaces such that they can be used in different locales.
+
+## Translation
+
+To support the multi-lingual translation of text in our interfaces, use the [LocalizeMixin](https://github.com/BrightspaceUI/core/tree/main/mixins/localize).
```
dlockhart, line 16: I'd love to be able to embed this page from core right inline here, but that isn't something Daylight supports right now.

pages/fundamentals/internationalization.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Numbers, Dates and Times
+
+When parsing or displaying numbers, dates, times or file sizes, use the [@brightspace-ui/intl library](https://github.com/Brightspaceui/intl). It will handle any necessary conversion to a user's locale and formatting preferences.
+
+## Bidirectionality & Right-to-Left (RTL) Languages
```
dlockhart, line 22: This is mostly borrowed from the now-deleted page from Style Elements, but I removed a big section on how numbers work in RTL. While interesting, none of it actually applied to us.

oelhanafey (commented):
I like the name fundamentals! I think it makes a lot of sense for this kind of stuff. 

Adding a new top level heading though is also going to add to the confusion when navigating. The `Designing` section is pretty bare bones and has a confusing name since `Style Elements` is probably where most designers want to be. If we are moving the accessibility page, I wonder if we can just compile the rest into a `Design Fundamentals` page and do away with that heading altogether.

pages/fundamentals/index.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: layouts/simple-content
+eleventyNavigation:
+  title: Fundamentals
+  key: fundamentals
+  order: 1
+---
+
```
dlockhart, line 8: Starting a conversation inline here so it can be threaded.

From Oscar:
> Adding a new top level heading though is also going to add to the confusion when navigating. The Designing section is pretty bare bones and has a confusing name since Style Elements is probably where most designers want to be. If we are moving the accessibility page, I wonder if we can just compile the rest into a Design Fundamentals page and do away with that heading altogether.

Yeah I was trying not to re-organization absolutely everything. Again comparing to Shopify, I think everything in "style elements" could go under "design"... typography, icons, illustrations, writing all seems "design-y" to me. The trick is that there is some technical content mixed in there, which is why it likely got pulled out.

I also know we originally landed where we are in terms of information architecture after a lot of discussion and voting and mind mapping. FWIW, Polaris has 11 top-level sections compared to our 8.

dlockhart:
Closing in favour of #1343.

PR #1320, opened 2023-04-25
Upgrade to rollup version 3

svanherk:
Decided to try moving to rollup version 3 while I was testing out the other dependency changes.  It _seems_ like this works, but I did a light regression test of functionality.  Looks like the tests are happy, but might be good for someone else to poke around too?


rollup.config.js
```diff
@@ -2,9 +2,11 @@ import { babel } from '@rollup/plugin-babel';
 import dynamicImportVars from '@rollup/plugin-dynamic-import-vars';
 import { default as files } from './.generated/rollup-files-generated-page.js';
 import { importMetaAssets } from '@web/rollup-plugin-import-meta-assets';
-import outputManifest from 'rollup-plugin-output-manifest';
+import pluginManifest from 'rollup-plugin-output-manifest';
 import resolve from '@rollup/plugin-node-resolve';
-import { terser } from 'rollup-plugin-terser';
+import terser from '@rollup/plugin-terser';
+
+const outputManifest = pluginManifest.default;
```
oelhanafey, line 9: Was this necessary because of the update? From [their docs](https://github.com/shuizhongyueming/rollup-plugin-output-manifest/tree/master/packages/main#usage) it looks like we were using it correctly

svanherk: Yeah, without it I was getting this error: https://github.com/shuizhongyueming/rollup-plugin-output-manifest/issues/21

oelhanafey (approved):
Looks good, the component demos would likely be broken if there was an issue with this

PR #1354, opened 2023-05-31
feat: rendering properties and code snippet separately

GZolla:
Add `demo properties` to parser in order to be able to show the live properties without being forced to show the demo snippet as well with `demo live`. This can be useful in cases like `d2l-summary-item` where to demo example is complex and is not necessary to show the code behind it.


.eleventy.cjs
```diff
@@ -236,18 +236,15 @@ module.exports = function(eleventyConfig) {
 			if (configs.name) openingTag += ` tag-name="${configs.name}" `;
 			if (configs.size) openingTag += ` size="${configs.size}" `;
 
-			if (content.includes('<!-- docs: demo live')) {
-
-				if (!configs.name) return `${openingTag}>`;
-
+			if ((content.includes('<!-- docs: demo live') && configs.name) || content.includes('<!-- docs: demo properties')) {
```
oelhanafey, line 239: Have you tested this by giving `demo properties` with no component name? I'm guessing that `configs.name` here is the component name that you add that the property relates to ([for example](https://github.com/BrightspaceUI/core/blob/5a84452f9fda107b393fa0d033a7816114e1a4f2/components/empty-state/README.md?plain=1#L67)). So if that's the case, I think we might also want this check for  `configs.name` on `demo properties` as well.

The site won't be able to generate the property list for a component if it isn't given the name.

oelhanafey (commented):
We can also document the new option [here](https://github.com/BrightspaceUI/documentation/blob/49636f4b0c7249395cd323dd06dae367f93beb3f/pages/about/adding-component.md?plain=1#L83)

oelhanafey (approved):

PR #1384, opened 2023-06-13
feat: update documentation properties

GZolla:
Update code block documentation files to use the new `code` and `properties` options.

Rally: [US153783](https://rally1.rallydev.com/#/?detail=/userstory/704396274525&fdp=true)


dlockhart (approved):
Nice, much clearer!

PR #1439, opened 2023-08-22
update event best practices

dlockhart:
I got asked whether event handlers on the host should be added in the constructor or `connectedCallback` and was surprised to see this recommend `firstUpdated`. I think we always want folks to use `connectedCallback`, ya?


pages/developing/best-practices/events.md
```diff
@@ -10,46 +10,46 @@ eleventyNavigation:
 # Events
 
 ## Component Architecture
-When your component communicates and interacts with other elements, the pattern we want to use is:
+When a component communicates and interacts with other elements, the pattern is:
```
dlockhart, line 13: Just fixing some you/yours while I'm in here as this seems a little simpler and less informal.

dbatiste:
> I think we always want folks to use `connectedCallback`, ya?

No, I don't think so. I think our rationale for wiring handlers up to the host is that they will be cleaned up automatically when the element is garbage collected, and we don't need to worry about adding/removing the handler if the element is removed and then re-added to the DOM. 

As for `firstUpdated` vs `constructor` - I think our thinking was that `firstUpdated()` would delay that wire-up work. Lit documentation mentions that event wire-up is extremely fast and provides examples wiring up in the constructor, however they also mention that if optimizing performance, the wire-up may be deferred using `firstUpdated` (or even later), but their example includes a further delay using `await new Promise((r) => setTimeout(r, 0));` in that case.

The only time wiring event handlers in `connectedCallback` is needed is when we need to manage the clean-up ourselves with `disconnectedCallback`, for example when we're wiring handlers to elements outside the elements context (ex. the `body`).

dlockhart:
Hmm, interesting. I guess the risk with `firstUpdated` though is that you also need to make sure you DO NOT remove it in `disconnectedCallback` or else it'll work the first time but stop working if the element is removed and then re-added to the DOM.

We're certainly not consistent with this in core from what I can tell.

dbatiste:
That's correct. It's one or the other.

dlockhart:
Ok thanks! I've updated this to be a bit more clear about the 3 different cases and the reasoning behind them.

dbatiste (approved):

PR #1448, opened 2023-09-01
use the new live infrastructure for PR previews

dlockhart:
Corresponding change to match [this change to the routes](https://github.com/Brightspace/d2l.dev/pull/4672), this uses the "live" infrastructure for the PR previews. I'll follow up afterwards to update the main release build.

Also switching to use `publish-to-s3` and `comment-on-pr` actions to better align with core and the blog.


.github/workflows/ci.yml
```diff
@@ -39,40 +38,27 @@ jobs:
...
-              issue_number: context.issue.number,
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              head_sha: "${{ github.event.pull_request.head.sha }}",
-              name: 'Preview URL',
-              status: 'completed',
-              conclusion: 'neutral',
-              output: { title: url, summary: `Preview your changes at: ${url}`}
-            })
+          bucket-path: s3://d2l.dev/live/prs/${{ github.repository }}/pr-${{ github.event.number }}/playground
```
dlockhart, line 52: Not _totally_ sure this will work... publishing the main site to `pr-1234/` and then the playground to a nested directory underneath. The alternative was `pr-1234/site` and `pr-1234/playground`, but then to visit the main site you'd need to go to `https://live.d2l.dev/prs/BrightspaceUI/documentation/pr-1234/site/`

pages/_includes/layouts/home-redirect.njk
```diff
@@ -1,3 +1,3 @@
 <head>
-  <meta http-equiv="Refresh" content="0; url='/about/overview/'" />
+  <meta http-equiv="Refresh" content="0; url='./about/overview/'" />
```
dlockhart, line 2: Otherwise visiting the preview redirects to `live.d2l.dev/about/overview/` which... doesn't exist.

.github/workflows/ci.yml
svanherk: Might as well try it

.github/workflows/ci.yml
```diff
@@ -39,40 +38,27 @@ jobs:
...
+          publish-directory: ./playground-build/
+      - name: Notify
```
svanherk, line 54: Do you want this posting every time?

dlockhart: Nope! Fixed.

.github/workflows/ci.yml
```diff
@@ -39,40 +38,28 @@ jobs:
...
           aws-session-token: ${{ secrets.AWS_SESSION_TOKEN }}
-          role-to-assume: arn:aws:iam::037018655140:role/daylight.d2l.dev-githubactions-access
+          role-to-assume: arn:aws:iam::022062736489:role/r+BrightspaceUI+documentation+repo
           role-duration-seconds: 3600
           aws-region: ca-central-1
       - name: Deploy - branch
-        run: |
-          find ./_site/ \( -iname '*.html' -o -iname '*.js' -o -iname '*.json' -o -iname '*.svg' \) -exec brotli {} \; -exec mv {}.br {} \;
-          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'br' --exclude '*.*' --include '*.html' --include '*.json' --include '*.svg'
-          aws s3 sync --delete ./_site/ "s3://daylight.d2l.dev/branches/pr-${{ github.event.number }}" --content-encoding 'br' --cache-control "public,max-age=31536000,immutable" --exclude '*.*' --include '*.js'
```
svanherk, line 49: Seems like we had some caching here before but now we don't.  Is that actually better though, if it's being replaced every push?

dlockhart: Sorry meant to mention that -- I intentionally didn't enable `publish-to-s3`'s caching for that exact reason.

svanherk:
This looks good to me other than the fact it's broken! 😝

margaree:
I'm going to be experimenting with some changes to try to get this working, so ignore for now

margaree:
I think this is ready for an initial review. The site generally works including demos and side nav. I have minimally corrected the hardcoded images and links but was wanting to see if this looks so far so good before I go too far down that road, but there are examples of how that would be done in `base.njk`.

dlockhart (commented):
Approach looks good to me!

margaree:
Ignore for now - I'm going through and testing each page. There's a few things that still need fixing up (github logo image, displaying labs page images, and removing what's new)

margaree (commented):
This is ready for review. Sorry for how giant it is!

How it works:
- `ASSET_PATH` gets set in CI to `/prs/${{ github.repository }}/pr-${{ github.event.number }}` (this env var can also be used if anything else moves like this in the future)
- `pages/_data/project.cjs` (which is where env vars can be taken and processed for use in templates/markdown pages) sets `project.assetPath` to be either the `ASSET_PATH` env var or empty string
- markdown and njk (nunjuck template) files use `{{ project.assetPath }}` to prefix any links
- components that have links generally have a property called `assetPath` which is used to prefix the links (this part gets a bit messy - I tried to go off the assumption that `assetPath` could be `undefined` in components even though many usages where `{{ project.assetPath }}` is used do just set it to empty string)

In theory, as long as the component code is solid, anything that's incorrect should only affect the CI builds, since if prefixes are added they default to empty string, which is what's used locally and in prod.

dlockhart (commented):
Haha I can't approve this because I'm the OG author, but looks good!

margaree (approved):

PR #1534, opened 2024-01-04
US159909 Update checkbox-drawer repo link

ChrisLabattD2L:
Context: [US159909](https://rally1.rallydev.com/#/?detail=/userstory/730793441071&fdp=true): Migrate Checkbox-Drawer to the Labs Monorepo

With the [`checkbox-drawer`](https://github.com/BrightspaceUILabs/checkbox-drawer) component getting moved over to the [monorepo](https://github.com/BrightspaceUI/labs/tree/main/src/components/checkbox-drawer), the repo link needs to be updated.

While making the changes, I also noticed that the `checkbox-drawer` component page cannot be access unless you're searching for it (i.e. it doesn't appear under the `Layout & Structure` section), so I'm wondering if there's any requirement that needs to be met for it to appear there like other labs components, or if I'm free to add it here as well.


margaree (approved):
Looks good! We don't actually want all components being displayed in the side nav. We just prioritized a select few that were most often used. It does show up on the status page (https://live.d2l.dev/prs/BrightspaceUI/documentation/pr-1534/components/status/#labs-components). This is likely something we should document more though, so thanks for mentioning that.

We (probably Gaudi) likely will want to improve the look of the link though as it's quite lengthy, but that's separate from this.
Screenshot 2024-01-04 at 11 32 09 AM

ChrisLabattD2L:
> We don't actually want all labs components being displayed in the side nav. We just prioritized a select few that were most often used. 

Gotcha, that makes sense. Thanks Margaree!

PR #1546, opened 2024-01-25
US159933 Update labs pagination component to point to monorepo

ChrisLabattD2L:
### Context for [US159933](https://rally1.rallydev.com/#/?detail=/userstory/730840760195&fdp=true): Update Pagination Dependencies

Now that the [`d2l-labs-pagination`](https://github.com/BrightspaceUILabs/pagination) component has been renamed (to `d2l-labs-pager-numeric`) and moved over to the [labs monorepo](https://github.com/BrightspaceUI/labs/blob/main/src/components/pagination), I'm updating the documentation here so that it points to the new code instead.


svanherk (approved):
The PR description is a bit confusing because it points to the `core` load more component 😛, but the actual code changes here look good!

ChrisLabattD2L:
> The PR description is a bit confusing because it points to the `core` load more component 😛, but the actual code changes here look good!

Oof, don't know how I managed to do things so incorrectly 😓 Glad I got the code right at least 😆

PR #1547, opened 2024-01-25
Add labs monorepo to documentation, update pager-numeric to point to labs monorepo and use that README

margaree:


tools/component-pages-data.js
```diff
@@ -184,28 +184,29 @@ export const componentPagesData = {
...
-				'layout': 'layouts/component-issue',
+				'layout': 'layouts/component',
 				'title': 'Pagination',
-				'fileName': 'pagination',
-				'repo': 'https://github.com/BrightspaceUILabs/pagination/',
-				'keywords': [
-					'paging',
-					'numeric'
-				],
+				'fileName': 'pager-numeric',
```
margaree, line 189: This got changed to `pager-numeric` but I can change it back to `pagination` to keep links working (or we can update it to something else?)

tools/component-pages-data.js
```diff
@@ -184,28 +184,29 @@ export const componentPagesData = {
...
 			'info': {
+				'baseInstallLocation': '@brightspace-ui/labs',
 				'published': true,
-				'internal': false
-			},
-			'content': '\r\n\r\n**If possible, use the new [Paging component](../../components/paging) instead of this labs component.**\r\n\r\nA component to indicate the existence of, and provide navigation for, multiple pages of content.\r\n\r\n<img width="647" alt="Pagination component" src="{{ project.assetPath }}/img/screenshots/pagination.png">\r\n\r\nComponents:\r\n- `d2l-labs-pagination`'
+				'internal': false,
+				'demoComponentImports': [
+					'components/pager-numeric.js'
+				]
```
margaree, line 208: This piece would allow live demos to work. If we aren't going to add those in yet we can remove this at this point and add it later.

PR #1548, opened 2024-01-25
Add labs monorepo, update pagination page to use labs monorepo page

margaree:
[Pagination Page](https://live.d2l.dev/prs/BrightspaceUI/documentation/pr-1548/components/pagination/)


ChrisLabattD2L (commented):
Would it be possible to also do this for the [`checkbox-drawer`](https://github.com/BrightspaceUI/labs/tree/main/src/components/checkbox-drawer) component that was also recently moved over to the monorepo?

tools/generate-pages.js
```diff
@@ -64,7 +64,7 @@ function _copyCustomElements(repos) {
 		const customElementsFilePath = path.join(__dirname, `../node_modules/${repo}/custom-elements.json`);
 		if (!existsSync(customElementsFilePath)) {
 			const message = `custom-elements.json does not exist for ${repo}`;
-			_handleError(message);
+			console.warn(`WARNING: ${message}`);
```
margaree, line 67: This was failing for the prod build because the labs monorepo does not yet have `custom-elements.json` being generated. I don't think this is something we should fail for - to me it's totally valid that a page could be pulled into Daylight that doesn't have live demos and therefore doesn't need `custom-elements.json`. They would hit problems if they tried to add the live demos and didn't have it, but generating it [is documented in Daylight](https://daylight.d2l.dev/about/adding-component/#generating-custom-elements.json).

I don't remember though if there was another reason why we failed on this scenario so let me know if I'm missing something.

tools/component-pages-data.js
```diff
@@ -261,19 +263,21 @@ export const componentPagesData = {
 		},
 		{
 			'frontMatter': {
-				'layout': 'layouts/component-issue',
+				'layout': 'layouts/component',
 				'title': 'Checkbox Drawer',
 				'fileName': 'checkbox-drawer',
```
margaree, line 268: Note that checkbox drawer doesn't have an entry in the side nav. Since we're adding it to the monorepo and it's getting a generated page, should it? @geurts

margaree: That's true. The search algorithm could definitely use another look. It can be found on [the status page](https://daylight.d2l.dev/components/status/).

ChrisLabattD2L (approved):

PR #1558, opened 2024-01-31
Remove accessible-toaster page

margaree:


tools/component-pages-data.js
dlockhart, line 1: Wondering if we should just remove it entirely instead?

margaree: I'm okay with that

dlockhart (approved):

PR #1559, opened 2024-01-31
Search: replace special characters with spaces for searching (e.g., hyphens), fix properties not being included in search

margaree:
This fixes:
- Not being able to search for a component tag (e.g., d2l-button-subtle)
- Properties not showing up in search results (we actually have a bool at the top of `build-index` that is set to true, but if when this is in the wild it's too busy we can disable it)


pages/_components/search.js
```diff
@@ -293,7 +293,7 @@ export class DaylightSearch extends LitElement {
 	}
 
 	_search(searchInput) {
-		const splitSearch = searchInput.trim().split(' ');
+		const splitSearch = searchInput.trim().replace(/[^a-zA-Z0-9 ]/g, ' ').split(' '); // replace special characters with spaces (lunr doesn't handle special characters)
```
margaree, line 296: I think we can add a backslash before special characters instead but this was something I dug into months ago so I'd have to revisit.

tools/build-index.js
```diff
@@ -151,7 +151,7 @@ function parsePage(url, sidenavSection, $) {
 			if (el.tagName === 'd2l-daylight-demo-snippet') {
 				if (SHOW_PROPERTIES) {
 					const tagName = $(el).attr('tag-name');
-					if (!tagName || !componentDemoPageRecords || !componentDemoPageRecords[tagName]) return;
+					if (!tagName) return;
```
margaree, line 154: This was always returning because `componentDemoPageRecords[tagName]` would never return true

pages/_components/search.js
dlockhart, line 296: Did you intend to have the space after `9`?

margaree: 🤦‍♀️

dlockhart (approved):

PR #1581, opened 2024-02-27
fix: Update highlight colour on last row when edited

saifaldin14:
Now when the last row is edited the highlight colour will be applied to it like the other rows

https://github.com/BrightspaceUI/documentation/assets/24593380/d7175126-ef73-4e44-a6ee-16cb58104abb


margaree:
Any ideas what's going on with the visual diffs? Looks like the corner is no longer rounded. Also, would we still want some sort of gradient for this case or are we good with the solid colour?
It could be useful to add a new visual diff to capture this case as well.

test/visual-diff/demo-tables.visual-diff.html
```diff
@@ -302,5 +303,8 @@
 		<div class="visual-diff">
 			<d2l-daylight-demo-tables id="hide-slots" tag="d2l-real-component" hide-slots></d2l-daylight-demo-tables>
 		</div>
+		<div class="visual-diff">
+			<d2l-daylight-demo-tables id="advanced-interactive-edit" tag="d2l-real-component" interactive _editMode defaults ='{"button-text":"My Button"}'></d2l-daylight-demo-tables>
+		</div>
```
margaree, line 308: Since this is the same content as the `interactive-edit` table above, can you just use that one?

test/visual-diff/demo-tables.visual-diff.html
```diff
@@ -95,8 +95,9 @@
 					{
 						"name": "number-of-things",
 						"description": "How many things",
-						"type": "number"
-					}
+						"type": "boolean",
+						"default": "false"
+					},
```
margaree, line 100: Can you just swap this one with `hide-close-button` since that's already a bool?

margaree (approved):

PR #1589, opened 2024-03-05
Add localize-mixin-sandbox to Daylight site

bearfriend:
This is a shift from BrightspaceUI/core#3503 to add the sandbox to the Daylight site instead.

Additions since the last update there:
- Template notes give tips and context
- Pretty Print fixes up bad/ugly messages
- Dates, numbers, and skeletons
  - These can be disabled until we have custom formatters in `core`

I have separated out changes to the rest of the site to add general localization docs and updated testing docs so we can finally get this up for people to use.


bearfriend:
Yikes, I'll try to get the whitespace issues figured out.

